PR #1, opened 2017-03-21
Adding the Lock free version of the KCAS algorithm

fondation451:


kayceesrk:
Thanks @fondation451. Before we merge this, can you install the reagents (https://github.com/ocamllabs/reagents) library (available from multicore-opam) and test it with the new k-CAS implementation. If you build reagents locally, then it will install several test programs, which you can run locally. We don't have proper tests yet, which we should fix.

kayceesrk:
@fondation451 Is this PR ready to merge?

PR #9, opened 2021-06-10
Backoff with cpu_relax

### LINKED ISSUE #8: Port Backoff module to use `Domain.Sync.cpu_relax`
lockfree library already does something similar: https://github.com/ocaml-multicore/lockfree/blob/6846bf4b16bcc61c60ecad43d2de3941ca56b1bf/src/backoff.ml

https://github.com/ocaml-multicore/ocaml-multicore/pull/561 removes `Domain.Sync.{critical_section, wait_for}`, so this needs to be updated.

Sudha247:
Fixes #8. Based on a version of exponential backoff in the lockfree library.

### DIFF
--- a/src/backoff.ml
+++ b/src/backoff.ml
@@ -1,6 +1,7 @@
 (*
  * Copyright (c) 2015, Théo Laurent <theo.laurent@ens.fr>
  * Copyright (c) 2015, KC Sivaramakrishnan <sk826@cl.cam.ac.uk>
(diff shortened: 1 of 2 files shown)


ctk21:
This will spin vs the old more restful wait (you could imagine using `sleep` I guess). However it matches how we are handling this in lockfree and fixes the issue which makes `kcas` not compile. 

Merging.

PR #11, opened 2022-02-02
update cpu_relax to new location

jmid:
This PR updates `Domain.Sync.cpu_relax` to its new location, so that kcas (and lockfree which depends on kcas) compiles with ocaml trunk/5.00.


kayceesrk:
Thanks.

PR #4, opened 2019-04-11
First pass at getting this to build under dune

sadiqj:

PR #12, opened 2022-03-09
Clean up dependencies

jmid:
This PR cleans up the dependencies listed in the opam-file, leaving only dune, following #4.
The Makefile is updated and I've removed the old ocamlbuild .mllib-file.


kayceesrk:
Thanks.

PR #13, opened 2022-11-23
Correct opamfile

bartoszmodelski:
I want to push kcas into opam for use in reagents. This PR updates paths in opam file. 

(I'll mark it as experimental release).


kayceesrk:
Thanks!

PR #16, opened 2022-12-08
Reorganize tests

bartoszmodelski:
Based on (https://github.com/ocaml-multicore/kcas/pull/15).

Clean up existing tests: 
- asserts instead of log success fails
- remove waits to stabilise and speed up tests
- don't shadow stdlib 
- etc 

Also remove Obj.id trick  

Add a few tests that document currently broken behavior - kCAS on a list with multiple instances of the same atomic creates broken result. At the moment reagents also lack a check, so they would have broken state as well.


test/test_overlapping_loc.ml
```diff
@@ -0,0 +1,44 @@
...
+  *)
+  let v_1 = Kcas.ref 0 in
+  let cas_1 = Kcas.mk_cas v_1 0 1 in
+  let cas_2 = Kcas.mk_cas v_1 0 1 in
+  assert (Kcas.kCAS [ cas_1; cas_2 ]);
+  assert (Kcas.get v_1 == 1)
+
+let test_3 () =
+  (* [cas_2] acts on the same location as [cas_1].
+     This is not something that we can handel in the
```
tmcgilchrist, line 29:
```suggestion
     This is not something that we can handle in the
```

test/test.ml
```diff
@@ -11,205 +11,128 @@
...
+  let a2 = Kcas.ref 0 in
+
+  let domains =
+    [
+      (fun () -> thread1 (a1, a2));
+      (fun () -> thread2 (a1, a2));
+      (fun () -> thread3 (a1, a2));
+    ]
+  in
+  List.map Domain.spawn domains |> List.iter Domain.join
```
polytypic, line 92: Again, I realise this has not fundamentally changed in this PR, but this test has a race condition.  It is (theoretically) possible that each domain gets spawned and completes before other domains.

polytypic: There are many ways to fix this.

What I might suggest is to make it so that `thread2` and `thread3` are changed to wait for a condition A before starting the interference loop.  Instead of performing a fixed amount of iterations, `thread2` and `thread3` would run their loops until another condition B becomes true.  `thread1` then first makes the condition A true and then runs the loop after which it makes condition B true.  Additionally `thread2` and `thread3` could increment counters that `thread1` checks periodically, perhaps two nested loops in `thread1`, for having been actually incremented during the inner loop to check that `thread2` and `thread3` are actually running during the inner loop in `thread1`.

bartoszmodelski:
Thanks for the suggestions @polytypic! I've addressed both.

test/test.ml
```diff
@@ -24,140 +24,154 @@
...
-
-  Domain.spawn (fun () -> thread4 (a1, a2)) |> ignore;
-  Domain.spawn (fun () -> thread5 (a1, a2)) |> ignore;
+  let a1 = Kcas.ref 0 in
+  let a2 = Kcas.ref 0 in
 
-  Unix.sleep wait_time;
-  !th4_success && !th5_success
+  let domains =
+    [ (fun () -> thread4 (a1, a2)); (fun () -> thread5 (a1, a2)) ]
```
polytypic, line 137: Does this also allow the possibility that one of the two domains completes before the other is started?

bartoszmodelski: fixed this one as well

PR #17, opened 2022-12-08
Detect overlap

bartoszmodelski:
Based on (https://github.com/ocaml-multicore/kcas/pull/16). 

Add location overlap detection. Currently, if list with cases contains the same location multiple times, undefined behavior ensues. Let's detect that and fail explicitly. 

This is related to swap_test2 in reagents, which hangs on impossible reagent. 

The check kicks in only if there's an overlap or another thread progressed the transaction for us (in which case the extra overhead probably doesn't matter much).


test/test_overlapping_loc.ml
```diff
@@ -0,0 +1,43 @@
+let test_1 () =
+  (* [cas_2] acts on the same location as [cas_1]
+     and has conflicting values. kCAS should fail.
+  *)
+  let v_1 = Kcas.ref 0 in
+  let cas_1 = Kcas.mk_cas v_1 0 1 in
+  let cas_2 = Kcas.mk_cas v_1 2 3 in
+
+  match Kcas.kCAS [ cas_1; cas_2 ] with exception _ -> () | _ -> assert false
```
polytypic, line 9: This test (and `test_3` below) makes an assumption that the cas ops in the list given to kCAS are performed sequentially in the list order.  If the algorithm decides to try the `cas_2` first, the whole kCAS fails and returns false before detecting overlap.  I believe it would be best to avoid that assumption and allow the algorithm to choose whichever order it wants.  One way to avoid the assumption would be to run kCAS twice with both orders (`Kcas.kCAS [ cas_1; cas_2 ] || Kcas.kCAS [ cas_2; cas_1 ]`) and assert that one of them raises an exception (the other might just return false).

test/test.ml
```diff
@@ -11,205 +11,128 @@
...
   done;
-  ignore @@ kCAS c1;
-  ()
-;;
+  ignore @@ Kcas.kCAS c1
 
 let thread2 (a1, a2) =
-  let c1 = [mk_cas a1 v_y v_x ; mk_cas a2 v_x v_y] in
-  let c2 = [mk_cas a1 v_x v_y ; mk_cas a2 v_y v_x] in
+  let c1 = [ Kcas.mk_cas a1 1 0; Kcas.mk_cas a2 0 1 ] in
```
polytypic, line 49: It seems that if the order of cas operations in the list is reversed
```diff
-  let c1 = [ Kcas.mk_cas a1 1 0; Kcas.mk_cas a2 0 1 ] in
+  let c1 = [ Kcas.mk_cas a2 0 1; Kcas.mk_cas a1 1 0 ] in
```
the test gets stuck.  I could imagine a library like kcas where that would be by design (IOW, the user is required to order the cas operations consistently).  Is this is a bug or not?

bartoszmodelski: Not a bug as long as we consider this library a faithful implementation of the kcas paper, which expects all casns to follow the same total order.  
>The update addresses lie in the data section of memory and are held according to some total order agreed by all threads to guarantee non-blocking behaviour,

But yeah, I did not realize, even though it seems logical in hindsight. I'll add a check for this.

test/test_overlapping_loc.ml
bartoszmodelski: Yeah, that's not ideal; moving the check to the beginning of kCAS function should fix it.

src/kcas.ml
```diff
@@ -81,39 +77,56 @@ let rec rdcss_read a =
...
-              phase1 curr_c_t_tail curr_st (CAS (a, o, n) :: out)
-            else phase1 curr_c_t_tail FAILED (CAS (a, o, n) :: out) )
+              ignore @@ casn_proceed c';
+              phase1 curr_cas_list curr_status out)
+            else (
+              (* suspicious - either another thread finished for
+                 us or there's an overlapping kcas *)
+              if
+                Option.is_some
+                  (List.find_opt (fun (CAS (v, _, _)) -> v.id = atomic.id) out)
```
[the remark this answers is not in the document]
bartoszmodelski: Yeah, definitely could degrade to quadratic in some cases. I'll change it to an explicit check at the beginning of kCAS.

bartoszmodelski:
I've added sorting of CASes at the beginning of kCAS and moved overlap detection there. It's not quite O(n) as it used to be but we get a lot of safety relatively cheaply.

polytypic:
> I've added sorting of CASes at the beginning of kCAS and moved overlap detection there. It's not quite O(n) but we get a lot of safety relatively cheaply.

Hmm... I wonder if there should be two different versions of the `kCAS` function.  One that internally sorts the CASes and another that merely asserts that the `CASes` are sorted?

bartoszmodelski:
Added a toggle to disable automatic sorting. 

Also, on a second thought, disabling sorting disables pre-flight checks. I think it makes sense as any total order will do, and the one according to our ids does not have to be the preferable one, so we can just leave an overall best-performance hatch.

polytypic (approved):
LGTM

Issue #7, opened 2021-02-08
Switching to GKMZ algorithm?


clayrat:
It seems that the current implementation uses the Harris-Fraser RDCSS-based algorithm, however, there's a new [Guerraoui-Kogan-Marathe-Zablotchi](https://arxiv.org/abs/2008.02527) algorithm which claims to be more optimal by only requiring k+1 CASes instead of 3k+1.

PR #20, opened 2023-01-06
GKMZ adaptation

polytypic:
This implements an adaptation of the [GKMZ algorithm](https://github.com/ocaml-multicore/kcas/issues/7) to OCaml.

With the only benchmark in this repository, performance seems slightly worse at 1-CAS and superior at n-CAS (when n > 1):

```sh
➜  kcas git:(gkmz-adaptation) hyperfine "_build/default/test/benchmark.exe 1 10000"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Time (mean ± σ):       3.5 ms ±   0.2 ms    [User: 2.7 ms, System: 0.4 ms]
  Range (min … max):     3.1 ms …   4.0 ms    517 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
➜  kcas git:(gkmz-adaptation) hyperfine "_build/default/test/benchmark.exe 2 10000"
Benchmark 1: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       7.8 ms ±   0.2 ms    [User: 7.0 ms, System: 0.4 ms]
  Range (min … max):     7.4 ms …   8.6 ms    299 runs
 
➜  kcas git:(gkmz-adaptation) hyperfine "_build/default/test/benchmark.exe 4 10000"
Benchmark 1: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      12.6 ms ±   0.3 ms    [User: 11.7 ms, System: 0.4 ms]
  Range (min … max):    12.2 ms …  14.4 ms    199 runs
 ```
 
 ```sh
 ➜  kcas git:(detect-overlap) hyperfine "_build/default/test/benchmark.exe 1 10000"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Time (mean ± σ):       3.2 ms ±   0.2 ms    [User: 2.6 ms, System: 0.3 ms]
  Range (min … max):     2.8 ms …   3.8 ms    539 runs
 
  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
 
➜  kcas git:(detect-overlap) hyperfine "_build/default/test/benchmark.exe 2 10000"
Benchmark 1: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):      17.4 ms ±   0.2 ms    [User: 16.5 ms, System: 0.5 ms]
  Range (min … max):    16.9 ms …  18.2 ms    151 runs
 
➜  kcas git:(detect-overlap) hyperfine "_build/default/test/benchmark.exe 4 10000"
Benchmark 1: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      31.1 ms ±   0.4 ms    [User: 30.2 ms, System: 0.5 ms]
  Range (min … max):    30.4 ms …  32.5 ms    90 runs
```

This adaptation simplifies the data structures so that a ref always contains a CAS descriptor (a record) rather than possibly a value (requiring a sum type).

I tried to use an array for the intermediate data, but it requires n-1 more write barriers when preparing for the kCAS.  Allocating lists turned out to be faster.

I chose not to change the library interface in the PR, aside from require CASes to be in strictly ascending order, but I think the library could use some tuning.  Note that the previous implementations of this library also require the CASes to always be in strictly ascending or descending order.  If CASes are not ordered, the previous implementations of this library may also get stuck.  This implementation verifies the order and fails with an `Invalid_argument` exception if the ordering is violated.


src/kcas.ml
```diff
@@ -8,181 +8,167 @@ module type Backoff = Backoff.S
...
-
-let kCAS c_l =
-  match c_l with
+let set atom value =
+  Atomic.set atom.state { before = value; after = value; casn = casn_after }
+
+let rec release_after = function
+  | [] -> true
+  | CAS (_, cas) :: cass ->
+      cas.casn <- casn_after;
```
bartoszmodelski, line 45: The status has already been updated on line 65, do we need this one (and the one on line 52)?

src/kcas.ml
```diff
@@ -8,181 +8,167 @@ module type Backoff = Backoff.S
...
+let rec determine casn undetermined = function
+  | CAS (atom_state, state) :: cass_1 as cass_0 ->
+      let state' = Atomic.get atom_state in
+      if state == state' then determine casn undetermined cass_1
+      else
+        let before = state.before in
+        let before' = state'.before and after' = state'.after in
+        if
+          (before == before' && before == after')
+          || before == if is_after state'.casn then after' else before'
```
bartoszmodelski, line 78: This logic of checking whether previous operation finished and getting its result happens here, in `cas` and `get`. Reusing `get` would make it more readable.

src/kcas.ml
```diff
@@ -8,181 +8,167 @@ module type Backoff = Backoff.S
...
+
+let rec determine casn undetermined = function
+  | CAS (atom_state, state) :: cass_1 as cass_0 ->
+      let state' = Atomic.get atom_state in
+      if state == state' then determine casn undetermined cass_1
+      else
+        let before = state.before in
+        let before' = state'.before and after' = state'.after in
+        if
+          (before == before' && before == after')
```
bartoszmodelski, line 77: Do we lose linearization on "noop cas", i.e. one where before = after, since it looks done before the entire kCAS completes?

So for something like [cas a 0 0; cas a 0 1] and [cas a 0 1] both could succeed even if there was never a state a=0, b=1.

Maybe worth disallowing "noop" inputs?

bartoszmodelski (commented):
Thanks for this PR, pretty neat! 

I've added a few minor comments but other than that looks good to me

src/kcas.ml
bartoszmodelski: Amazingly, it seems to be an actual use-case: https://github.com/ocaml-multicore/lockfree/blob/a72d081e550b589cb3106f87d4b345f5e1b82470/src/paradict.ml#L132

polytypic: Note that the `before'` and `after'` values are read from the state of the atom rather than the input.

src/kcas.ml
polytypic: See the comment on line 56.  This reduces space (by freeing an atomic object) at the expense of a write barrier.

It is not entirely clear to me whether freeing the space is worth it.

I guess it boils down to usage patterns.  Speculation: If you constantly keep updating most kcas refs, then the freeing the space is probably not worth it.  If, OTOH, you have a lot of kcas refs and only update a few of those at any given moment, then it might be relatively more beneficial to free the space.

polytypic:
I made a quick experiment to use [`fenceless_get`](https://ocaml-multicore.github.io/multicore-magic/multicore-magic/Multicore_magic/index.html#val-fenceless_get).

`fenceless_get` is a workaround to avoid unnecessary acquire fences in lock-free algorithms in the current release of multicore OCaml.

Here is a set of benchmark runs *without* `fenceless_get`:

```sh
➜  kcas git:(gkmz-adaptation) ✗ P=_build/default/test/benchmark.exe; N=200000; hyperfine --warmup 1 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      34.7 ms ±   0.6 ms    [User: 33.6 ms, System: 0.7 ms]
  Range (min … max):    34.1 ms …  39.0 ms    76 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (39.0 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: _build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     107.2 ms ±   0.9 ms    [User: 105.8 ms, System: 1.0 ms]
  Range (min … max):   105.4 ms … 108.2 ms    28 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     195.3 ms ±   1.6 ms    [User: 193.7 ms, System: 1.2 ms]
  Range (min … max):   193.1 ms … 197.9 ms    15 runs
 
Summary
  '_build/default/test/benchmark.exe 1 200000' ran
    3.09 ± 0.06 times faster than '_build/default/test/benchmark.exe 2 200000'
    5.63 ± 0.10 times faster than '_build/default/test/benchmark.exe 4 200000'
```

Here is the same set *with* `fenceless_get`:

```sh
➜  kcas git:(gkmz-adaptation) ✗ P=_build/default/test/benchmark.exe; N=200000; hyperfine --warmup 1 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      20.4 ms ±   0.8 ms    [User: 19.4 ms, System: 0.7 ms]
  Range (min … max):    20.1 ms …  28.0 ms    107 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (28.0 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: _build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     104.6 ms ±   0.6 ms    [User: 103.3 ms, System: 0.9 ms]
  Range (min … max):   103.5 ms … 105.6 ms    28 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     178.1 ms ±   0.7 ms    [User: 176.3 ms, System: 1.3 ms]
  Range (min … max):   176.6 ms … 178.9 ms    16 runs
 
Summary
  '_build/default/test/benchmark.exe 1 200000' ran
    5.12 ± 0.20 times faster than '_build/default/test/benchmark.exe 2 200000'
    8.71 ± 0.33 times faster than '_build/default/test/benchmark.exe 4 200000'
```

polytypic:
I rebased this PR agains the `main`.

Here is new baseline on `main` that includes the new `presort` flag:

```sh
➜  kcas git:(main) P=_build/default/test/benchmark.exe; N=10000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Time (mean ± σ):       3.4 ms ±   0.1 ms    [User: 2.7 ms, System: 0.5 ms]
  Range (min … max):     3.4 ms …   4.2 ms    706 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.2 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):      20.3 ms ±   0.2 ms    [User: 19.3 ms, System: 0.7 ms]
  Range (min … max):    20.1 ms …  22.6 ms    147 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      38.9 ms ±   0.2 ms    [User: 37.8 ms, System: 0.7 ms]
  Range (min … max):    38.5 ms …  39.4 ms    77 runs
 
Summary
  '_build/default/test/benchmark.exe 1 10000' ran
    5.93 ± 0.14 times faster than '_build/default/test/benchmark.exe 2 10000'
   11.36 ± 0.24 times faster than '_build/default/test/benchmark.exe 4 10000'
```

Here is the GKMZ adaptation that also includes the new `presort` flag:

```sh
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=10000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.5 ms    673 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.5 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       9.6 ms ±   0.1 ms    [User: 8.7 ms, System: 0.6 ms]
  Range (min … max):     9.5 ms …   9.9 ms    311 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      18.9 ms ±   0.1 ms    [User: 18.0 ms, System: 0.6 ms]
  Range (min … max):    18.8 ms …  19.2 ms    157 runs
 
Summary
  '_build/default/test/benchmark.exe 1 10000' ran
    2.74 ± 0.06 times faster than '_build/default/test/benchmark.exe 2 10000'
    5.39 ± 0.12 times faster than '_build/default/test/benchmark.exe 4 10000'
```

Unexpectedly the presort step adds quite a bit of overhead.

polytypic:
I implemented an experiment to insert the CASes into a splay tree during preparation.  This ignores the `presort` flag.

Before (using `List.sort`):

```sh
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=10000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   3.8 ms    797 runs
 
Benchmark 2: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       9.6 ms ±   0.1 ms    [User: 8.7 ms, System: 0.6 ms]
  Range (min … max):     9.5 ms …   9.9 ms    308 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      18.9 ms ±   0.1 ms    [User: 18.0 ms, System: 0.6 ms]
  Range (min … max):    18.8 ms …  19.2 ms    158 runs
 
Summary
  '_build/default/test/benchmark.exe 1 10000' ran
    2.75 ± 0.04 times faster than '_build/default/test/benchmark.exe 2 10000'
    5.42 ± 0.07 times faster than '_build/default/test/benchmark.exe 4 10000'
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=200000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      35.1 ms ±   0.2 ms    [User: 34.1 ms, System: 0.7 ms]
  Range (min … max):    34.7 ms …  35.6 ms    85 runs
 
Benchmark 2: _build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     158.6 ms ±   1.4 ms    [User: 157.1 ms, System: 1.1 ms]
  Range (min … max):   156.5 ms … 161.2 ms    19 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     346.3 ms ±   2.8 ms    [User: 344.2 ms, System: 1.6 ms]
  Range (min … max):   341.2 ms … 349.2 ms    10 runs
 
Summary
  '_build/default/test/benchmark.exe 1 200000' ran
    4.52 ± 0.04 times faster than '_build/default/test/benchmark.exe 2 200000'
    9.86 ± 0.09 times faster than '_build/default/test/benchmark.exe 4 200000'
```

After (using a splay tree):

```sh
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=10000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.4 ms    688 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.4 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       9.4 ms ±   0.1 ms    [User: 8.5 ms, System: 0.6 ms]
  Range (min … max):     9.3 ms …   9.8 ms    316 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      15.6 ms ±   0.1 ms    [User: 14.7 ms, System: 0.6 ms]
  Range (min … max):    15.5 ms …  16.1 ms    191 runs
 
Summary
  '_build/default/test/benchmark.exe 1 10000' ran
    2.71 ± 0.06 times faster than '_build/default/test/benchmark.exe 2 10000'
    4.49 ± 0.10 times faster than '_build/default/test/benchmark.exe 4 10000'
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=200000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      34.5 ms ±   0.3 ms    [User: 33.4 ms, System: 0.7 ms]
  Range (min … max):    33.6 ms …  35.0 ms    87 runs
 
Benchmark 2: _build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     155.6 ms ±   1.4 ms    [User: 154.1 ms, System: 1.1 ms]
  Range (min … max):   153.1 ms … 157.8 ms    19 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     280.0 ms ±   1.3 ms    [User: 278.0 ms, System: 1.4 ms]
  Range (min … max):   277.8 ms … 281.4 ms    10 runs
 
Summary
  '_build/default/test/benchmark.exe 1 200000' ran
    4.52 ± 0.05 times faster than '_build/default/test/benchmark.exe 2 200000'
    8.12 ± 0.08 times faster than '_build/default/test/benchmark.exe 4 200000'
```

polytypic:
I tweaked the splay tree based implementation with a couple of optimizations.

```sh
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=10000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.0 ms    789 runs
 
Benchmark 2: _build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       8.0 ms ±   0.0 ms    [User: 7.1 ms, System: 0.6 ms]
  Range (min … max):     7.9 ms …   8.3 ms    375 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      13.3 ms ±   0.1 ms    [User: 12.3 ms, System: 0.6 ms]
  Range (min … max):    13.1 ms …  14.7 ms    226 runs
 
Summary
  '_build/default/test/benchmark.exe 1 10000' ran
    2.31 ± 0.04 times faster than '_build/default/test/benchmark.exe 2 10000'
    3.83 ± 0.07 times faster than '_build/default/test/benchmark.exe 4 10000'
```

```sh
➜  kcas git:(gkmz-adaptation) P=_build/default/test/benchmark.exe; N=200000; hyperfine --warmup 10 --shell none "$P 1 $N" "$P 2 $N" "$P 4 $N"
Benchmark 1: _build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      34.3 ms ±   0.3 ms    [User: 33.3 ms, System: 0.7 ms]
  Range (min … max):    33.6 ms …  34.8 ms    89 runs
 
Benchmark 2: _build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     126.1 ms ±   1.0 ms    [User: 124.7 ms, System: 1.0 ms]
  Range (min … max):   124.6 ms … 127.1 ms    24 runs
 
Benchmark 3: _build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     230.9 ms ±   2.0 ms    [User: 229.1 ms, System: 1.3 ms]
  Range (min … max):   229.2 ms … 234.0 ms    12 runs
 
Summary
  '_build/default/test/benchmark.exe 1 200000' ran
    3.68 ± 0.04 times faster than '_build/default/test/benchmark.exe 2 200000'
    6.73 ± 0.08 times faster than '_build/default/test/benchmark.exe 4 200000'
```

polytypic:
Note that with the splay tree based implementation, a presorted cas list (either ascending or descending) gives linear time (guaranteed).  One results in a `/` shaped tree and the other in a `\` shaped tree.  Both perform linearly, but the `\` shaped tree uses less stack.

If user has some other total order that is not based on the `id` values in refs, then the splay tree linear case does not apply.

More on [splay trees](https://en.wikipedia.org/wiki/Splay_tree).

src/kcas.ml
polytypic: Yes. I believe I wrote it that way for minor performance reasons.  I'll try and see if I can reduce the duplication without giving up performance.

* [ ] check and report on whether duplication can be reduced

src/kcas.ml
polytypic: > So for something like `[cas a 0 0; cas b 0 1]` and `[cas a 0 1]` both could succeed even if there was never a state a=0, b=1.

Hmm... Took me a moment to think about what you might mean here.

At the start of kcas a new `state` record is allocated for each ref (or individual cas) in the operation and then those states are installed to the refs.  For a kcas to succeed, all of the states need to be installed and the status updated to `After`.  Which means that it is not possible for two operations to both succeed without the intermediate state.

polytypic:
Question: Would it be Ok to use `fenceless_get` and `fenceless_set` here?

src/kcas.ml
polytypic: Hmm... It seems to get tricky to avoid the duplication with or without performance degradation.

First, the existential poses typing challenges.  Moving `get` into the same `let rec` and calling it results in:

```
File "src/kcas.ml", line 78, characters 34-38:
78 |       else if state.before == get atom then
                                       ^^^^
Error: This expression has type $CAS_'a ref
       but an expression was expected of type 'a
       The type constructor $CAS_'a would escape its scope
```

This tends to happen when trying to extract pretty much any useful part out of the loop.

Second, there are some intricate choices made in the comparisons for performance.  In particular, calls to `is_after` and accessing the `casn` is delayed as much as possible. The seemingly redundant comparisons before the `is_after` call improve performance by not accessing the `casn` unless necessary and those comparisons slightly vary depending on what is at hand.  In particular,

bartoszmodelski (approved):
LGTM 

I've run the benchmarks on my x86; the performance was comparable for small values of n, and significantly better for high ones. Reagents tests pass with this branch pinned. Not a scientific comparison, but some of the reagents benchmarks seem 5-20% faster than before.

> Question: Would it be Ok to use `fenceless_get` and `fenceless_set` here?
> 
> _Addtion_: Let's ignore that for now. I'll probably create another PR to use fenceless operations. That PR can then serve as a case to consider for adding fenceless operations to OCaml and the topic can be discussed further in that PR.

Yep, I like this approach.

src/kcas.ml
```diff
@@ -13,165 +13,163 @@ module Id = struct
...
+  && Atomic.compare_and_set atom.state state' state
+
+let commit (T (r, expect, update)) = cas r expect update
+
+let get atom =
+  let state = Atomic.get atom.state in
+  let before = state.before and after = state.after in
+  if before == after || is_after state.casn then after else before
+
+let overlap () = failwith "kcas: location overlap" [@@inline never]
```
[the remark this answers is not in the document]
polytypic: The `overlap` function should never be called in a correct program.  Call of `overlap ()` rather than `failwith "..."` generates slightly shorter code on some platforms.  This potentially slightly reduces the cost of the `splay` function.  Probably doesn't make a big difference, but I have noticed that at least on Apple M1 code size/alignment may have large impact on performance.

test/test.ml
```diff
@@ -166,11 +166,55 @@ let test_stress n nb_loop =
...
+    let y = x + 1 in
+    Array.fold_left (fun cs r -> Kcas.mk_cas r x y :: cs) [] refs
+  in
+
+  let thread () =
+    let refs = Array.copy refs in
+    Random.self_init ();
+    Barrier.await barrier;
+    for _ = 1 to n_incs do
+      (* Tested that with [~presort:false] this likely hangs *)
```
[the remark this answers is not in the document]
polytypic: I added this as part of the commit that removes the `presort` flag.

polytypic:
Separate [PR to use fenceless operations](https://github.com/ocaml-multicore/kcas/pull/21).

bartoszmodelski:
(please ignore ocaml-benchmark fail - wip)

PR #22, opened 2023-01-31
Complete redesign of the kcas library API

polytypic:
This PR proposes a complete redesign of the API that divides the API into four modules `Backoff`, `Loc`, `Op`, and `Tx`.  The diffs show how to change client code to use the new API.

```diff
+module Loc = Kcas.Loc
+module Op = Kcas.Op
+module Tx = Kcas.Tx
```

As proposed in this PR, `Loc` is a module that has a signature that is essentially compatible with the signature of the stdlib `Atomic` module.  The main difference is that several operations take an optional `backoff` (and that there is a `Backoff` module). Client code wishing to perform multiple `compare_and_set` operations atomically could theoretically just switch to use `Loc` instead of `Atomic`.

```diff
-Atomic
+Loc
```

```diff
-Kcas.cas
+Loc.compare_and_set
```

The constructor function `ref` is also renamed to `make`.

```diff
-Kcas.ref initial
+Loc.make initial
```

The name`ref`, while not a keyword in OCaml, has a long history and it is probably better to avoid it.  The term "location" is used e.g. in the papers on the OCaml memory model.

The previous API contained two related operations named `try_map` and `map`.  I believe this is a mistake in a couple of ways.  First of all, the mnemonic `map` is generally used for functional updates while `kcas` is fundamentally imperative.  Second, the `map` operations had arguably cumbersome signatures and the signature of `map` was also imprecise.  This PR proposes a single simpler `update` operation that also requires fewer allocations.  Note how in this PR several other operations are implemented in terms of `update`.

```diff
-match Kcas.map r (fun c -> if condition c then None else Some (next c)) with
-| Success c -> (* success *)
-| Failed | Aborted -> (* aborted *)
+match Loc.update r (fun c -> if condition c then raise Exit else next c) with
+| c -> (* success *)
+| exception Exit -> (* aborted *)
```

```diff
-match Kcas.try_map r (fun c -> if condition c then None else Some (next c)) with
-| Success c -> (* success *)
-| Failed -> (* failed *)
-| Aborted -> (* aborted *)
+let c = Loc.get r in
+if condition c then (* aborted *)
+else if Loc.compare_and_set r c (next c) then (* success *)
+else (* failure *)
```

The previous API had a `W1` module.  That has been removed, because it didn't provide anything substantial beyond of what the stdlib `Atomic` already provides.

```diff
-Kcas.W1
+Atomic
```

Actually the `W1` module did provide `map` and `try_map`.  It would probably make sense to move the `Backoff` module to the stdlib and also to add an `Atomic.update` function.  I leave that to future work.

This PR also includes a redesign of the `Backoff` module to use an internal representation as a single immutable `int` and the `once` operation with `t -> t` type.  This way `Backoff` does not require memory allocations during a backoff loop and possibility of false sharing is also reduced. The default lower and upper backoff values are chosen to approximate what is currently in the `lockfree` library `Backoff` module.

The atypically named `kCAS` operation has been renamed to `atomically` and `commit` is now called `atomic`.  They are now inside the `Op` module.

```diff
-Kcas.kCAS
+Op.atomically
```

```diff
-Kcas.commit
+Op.atomic
```

```diff
-Kcas.mk_cas
+Op.mk_cas
```

```diff
-Kcas.is_on_ref
+Op.is_on_loc
```

This PR also adds an API as the `Tx` module for performing transactions on shared memory locations.  One may consider the transaction API as a higher-level alternative to constructing a list of operations as with the `Op` module. Transactions can be composed sequentially and conditionally.  This PR includes two different transactional queues and a stack and demonstrates that one can transfer elements between different data structures atomically.

Here is an example of committing a transaction that swaps the values of the two shared memory references `x_loc` and `y_loc` and returns their sum:

```ocaml
Tx.(
  commit begin
    let* x = get x_loc
    and* y = get y_loc in
    let+ () = set y_loc x
    and+ () = set x_loc y in
    x + y
  end
)
```

One potentially interesting avenue for further work would be to extend the algorithms and this library to support efficient compare-only operations.  One can express a kind of compare operation as `mk_cas loc expected expected`.  While that works, it writes to the location potentially causing contention and resulting in poor performance.  Efficient read-only transactions, for example, could be a useful addition to kcas.  As mentioned in

> [Nonblocking k-compare-single-swap](http://people.csail.mit.edu/shanir/publications/Nonblocking%20k-compare.pdf)
> Luchangco, Moir, Shavit

operation with multiple compares and a single swap also has uses.  I'll leave this to further work.


polytypic:
I made a quick test to build Reagents with the API in this PR:

```diff
modified   lib/offer.ml
@@ -41,9 +41,9 @@ module Make (Sched : Scheduler.S) : S = struct
     | Rescinded
     | Completed of 'a
 
-  type 'a t = 'a status Kcas.ref
+  type 'a t = 'a status Kcas.t
 
-  let make () = Kcas.ref Empty
+  let make () = Kcas.make Empty
   let get_id r = Offer_id.make (Kcas.get_id r)
   let equal o1 o2 = Kcas.get_id o1 = Kcas.get_id o2
 
@@ -54,22 +54,20 @@ module Make (Sched : Scheduler.S) : S = struct
 
   let wait r =
     Sched.suspend (fun k ->
-        let cas_result =
-          Kcas.map r (fun v ->
+        match
+          Kcas.update r (fun v ->
               match v with
-              | Empty -> Some (Waiting k)
+              | Empty -> Waiting k
               | Waiting _ | Catalyst -> failwith "Offer.wait(1)"
-              | Completed _ | Rescinded -> None)
-        in
-        match cas_result with
+              | Completed _ | Rescinded -> raise Exit)
+        with
         (* If CAS was a success, then it is no longer this thread's responsibiliy to
          * resume itself. *)
-        | Kcas.Success _ -> None
+        | _ -> None
         (* If the CAS failed, then another thread has already changed the offer from
          * [Empty] to [Completed] or [Rescinded]. In this case, thread shouldn't
          * wait. *)
-        | Kcas.Aborted -> Some ()
-        | Kcas.Failed -> failwith "Offer.wait(2)")
+        | exception Exit -> Some ())
 
   let complete r new_v =
     let old_v = Kcas.get r in
@@ -82,15 +80,14 @@ module Make (Sched : Scheduler.S) : S = struct
     | Rescinded | Completed _ -> PostCommitCas.return false (fun () -> ())
 
   let rescind r =
-    let cas_result =
-      Kcas.map r (fun v ->
-          match v with
-          | Empty | Waiting _ -> Some Rescinded
-          | Rescinded | Completed _ | Catalyst -> None)
-    in
-    (match cas_result with
-    | Kcas.Success (Waiting t) -> Sched.resume t ()
-    | _ -> ());
+    (match
+       Kcas.update r (fun v ->
+           match v with
+           | Empty | Waiting _ -> Rescinded
+           | Rescinded | Completed _ | Catalyst -> raise Exit)
+     with
+    | Waiting t -> Sched.resume t ()
+    | _ | (exception Exit) -> ());
     match Kcas.get r with
     | Rescinded | Catalyst -> None
     | Completed v -> Some v
@@ -101,7 +98,7 @@ module Make (Sched : Scheduler.S) : S = struct
   type catalyst = unit -> unit
 
   let make_catalyst () =
-    let offer = Kcas.ref Catalyst in
+    let offer = Kcas.make Catalyst in
     let cancel () = Kcas.set offer Rescinded in
     (offer, cancel)
 
modified   lib/postCommitCas.ml
@@ -14,26 +14,27 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-type 'a ref = 'a Kcas.ref
+type 'a ref = 'a Kcas.t
 
-let ref = Kcas.ref
+let ref = Kcas.make
 let get = Kcas.get
 
-type cas_kind = Real of Kcas.t | Imm of bool
+type cas_kind = Real of Kcas.op | Imm of bool
 type t = cas_kind * (unit -> unit)
 
 let cas r old_v new_v post_commit =
-  (Real (Kcas.mk_cas r old_v new_v), post_commit)
+  (Real (Kcas.CAS (r, old_v, new_v)), post_commit)
 
 let is_on_ref (c, _) r =
   match c with
   (* | Real cas -> Kcas.is_on_ref cas r *)
-  | Real cas -> Kcas.is_on_ref cas r
+  | Real (CAS (r', _, _)) -> Obj.magic r' == r
   | _ -> false
 
 let commit (cas, post_commit) =
   match cas with
-  | Real cas -> if Kcas.commit cas then Some post_commit else None
+  | Real (CAS (r, e, d)) ->
+      if Kcas.compare_and_set r e d then Some post_commit else None
   | Imm v -> if v then Some post_commit else None
 
 let return v post_commit = (Imm v, post_commit)
@@ -49,4 +50,4 @@ let kCAS lst =
       ([], (fun () -> ()), true)
       lst
   in
-  if live && Kcas.kCAS cas_list then Some post_commit else None
+  if live && Kcas.atomically cas_list then Some post_commit else None
modified   lib/postCommitCas.mli
@@ -14,7 +14,7 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-type 'a ref = 'a Kcas.ref
+type 'a ref = 'a Kcas.t
 
 val ref : 'a -> 'a ref
 val get : 'a ref -> 'a
modified   lib/ref.ml
@@ -41,7 +41,7 @@ module Make (Sched : Scheduler.S) :
   type mono_offer = Offer : 'a Offer.t -> mono_offer
 
   type 'a ref = {
-    data : 'a Kcas.ref;
+    data : 'a Kcas.t;
     offers : mono_offer Lockfree.Michael_scott_queue.t;
   }
 
@@ -50,7 +50,7 @@ module Make (Sched : Scheduler.S) :
   open Core
 
   let mk_ref v =
-    { data = Kcas.ref v; offers = Lockfree.Michael_scott_queue.create () }
+    { data = Kcas.make v; offers = Lockfree.Michael_scott_queue.create () }
 
   let rec read : 'a 'r. 'a ref -> ('a, 'r) reagent -> (unit, 'r) reagent =
    fun r k ->
@@ -83,7 +83,7 @@ module Make (Sched : Scheduler.S) :
       (fun (Offer offer) -> ignore (Option.is_none (Offer.rescind offer)))
       offers
 
-  let cas_imm r expect update = Kcas.cas r.data expect update
+  let cas_imm r expect update = Kcas.compare_and_set r.data expect update
 
   let rec upd :
             'a 'b 'c 'r.
@@ -100,7 +100,7 @@ module Make (Sched : Scheduler.S) :
         match f old_value arg with
         | None -> on_failure
         | Some (new_value, c) ->
-            if Kcas.cas ref.data old_value new_value then (
+            if Kcas.compare_and_set ref.data old_value new_value then (
               wake_all ref.offers;
               next_reagent.try_react c reaction offer)
             else Retry
```

polytypic:
The `Tx` mechanism is fundamentally limited as it does not support non-busy wait or blocking. Unfortunately, adding full support for blocking seems like it would be outside the scope of the underlying algorithms as it would add significant dependencies and/or significant overheads.

However, there might be practical ways to extend the `kcas` API to allow it to support low overhead blocking transactions on top of the underlying transaction log mechanism.

To support blocking, one essentially needs a way to signal waiters. After mutating some locations the mutator signals waiters. For a scalable mechanism that signal needs to be selective and only wake up those waiters that are interested in the mutated locations.

To associate waiters with locations in a truly low-overhead fashion, one possibility would be to allow locations to be "tagged":

```ocaml
module Loc : sig
  type ('tag, 'a) tagged

  val make_tagged: 'tag -> 'a -> ('tag, 'a) tagged
  val get_tag : ('tag, 'a) tagged -> 'tag

  type 'a t = (unit, 'a) t
  
  (* ... *)
```

In a blocking transaction mechanism that `'tag` could be a bag of the waiters of changes to the location.

Additionally, a scalable blocking mechanism also needs to be able to efficiently figure out which locations have been read and which have been written. A waiter needs to add itself to the read locations and a mutator needs to signal waiters of written locations.

```ocaml
module Tx : sig
  (* ... *)

  module Log : sig
    type t

    type 'r reducer = {
      one : 't 'a. ('t, 'a) Loc.tagged -> 'r;
      zero : 'r;
      plus : 'r -> 'r -> 'r;
    }

    val reduce : 'r reducer -> t -> 'r
    (** [reduce reducer] performs a fold over the transaction log. *)
  end

  exception Retry of unit t * (unit -> unit)
  (** Exception raised by {!reset_and_retry}. *)

  val reset_and_retry : read:(Log.t -> unit t) -> wait:(unit -> unit) -> 'a t
  (** [reset_and_retry read wait] returns a transaction that resets the
      current transaction such that it only reads from the accessed locations.
      The [read] function is then called with the internal transaction log to
      construct a transaction that is then composed after the current
      transaction.  The composed transaction [tx] and the given [wait] effect
      are then raised as a [Retry (tx, wait)] exception. *)

  val written: (Log.t -> unit t) -> 'a t -> 'a t
  (** [written on_written tx] returns a transaction that executes as [tx] and
      then calls the given function with a view of the internal transaction log
      restricted to the written locations.  The returned transaction is then
      composed after the transaction [tx].

      The intended use case for [written] is a to extend a transaction to signal
      waiters in a blocking transaction mechanism:

      {[
        let rec blocking_tx tx =
          let all_waiters = Loc.make [] in
          match
            tx
            |> written (fun log ->
                 (* remove all waiters of all written locations
                    and add them to the [all_waiters] list. *)
               )
            |> attempt
          with
          | result ->
            (* signal [all_waiters] *)
            result
          | exception Exit ->
            blocking_tx tx
          | exception Retry (add_waiters_tx, wait) -> (
            match attempt add_waiters_tx with
            | () -> wait ()
            | exception Exit ->
              (* Locations were already mutated before waiters could be added *)
              blocking_tx tx)
      ]} *)
```

The above is only a preliminary idea.  I have not yet fully implemented the above to verify it in practise.

bartoszmodelski (commented):
Thanks for the PR. The new interface is a lot cleaner. I left a few comments with questions and minor suggestions.

I experimented a bit with transactions and they look solid to me.

src/kcas.ml
```diff
@@ -95,134 +77,257 @@ and is_after casn =
...
+    [@@inline]
+
+  let get_as g loc (casn, cass) =
+    match cass with
+    | NIL -> get_as0 g loc casn NIL NIL
+    | _ -> (
+        match splay loc.id cass with
+        | l, Miss, r -> get_as0 g loc casn l r
+        | l, Hit (_loc', state'), r ->
+            (* assert (Obj.magic _loc' == loc); *)
```
bartoszmodelski, line 243:
```suggestion
```

bartoszmodelski:
This sounds good for API reference (with all other info being in the readme, eio-style). Do you have it up for the other project already?

polytypic:
> This sounds good for API reference (with all other info being in the readme, eio-style). Do you have it up for the other project already?

Yes and no.  I was planning to use it e.g. in `par-ml` and `idle-domains`, because they have multiple branches with different approaches.  I developed the scripts to do it, but I have yet to actually use it on those projects.

But the basic idea is very simple.  The script first clones gh-pages branch and clears it completely (locally).  It then iterates through all the git (branches or) tags, runs the commands to generate documentation, and if generation was successful, copies the generated documentation to the `gh-pages` clone under the name of the (branch or) tag.  So, you'll have a directory per doc version. To link to the documentation you just add the (branch or) tag name to the URL path:

```diff
-https://ocaml-multicore.github.io/kcas/doc/
+https://ocaml-multicore.github.io/kcas/main/
```

bartoszmodelski (approved):
LGTM

PR #24, opened 2023-02-09
New obstruction-free k-CAS-n-CMP algorithm

polytypic:
The core of the new k-CAS-n-CMP algorithm is described in the following gist:

> [Extending k-CAS with efficient read-only CMP operations](https://gist.github.com/polytypic/0efa0e2981d2a5fc4b534a0e25120cc9)

This is an extension of the previous GKMZ algorithm and preserves lock-free k-CAS as a subset of functionality.

---

This PR includes a simple benchmark with two transactions that read from shared locations, but only write to their own local locations.  With the algorithm in this PR, I get something like 100ns/tx.  With the previous algorithm timings vary considerably, but tend to be at least twice as much and often much more.  I suspect it is possible that with the previous algorithm the domains might not be able to both proceed at the same time and that would explain the peak performance of being only twice as slow.

### DIFF
--- a/src/kcas.ml
+++ b/src/kcas.ml
@@ -10,69 +10,101 @@ module Id = struct
   let get_unique () = Atomic.fetch_and_add id 1
 end
 
+type determined = [ `After | `Before ]
+
 type 'a loc = { state : 'a state Atomic.t; id : int }
 and 'a state = { mutable before : 'a; mutable after : 'a; mutable casn : casn }
 and cass = CASN : 'a loc * 'a state * cass * cass -> cass | NIL : cass
 and casn = status Atomic.t
-and status = [ `Undetermined of cass | `After | `Before ]
+and status = [ `Undetermined of cass | determined ]
+
+let is_cmp casn state = state.casn != casn [@@inline]
+
+module Mode = struct
+  type t = determined
+
+  let lock_free = (`After :> t)
+  let obstruction_free = (`Before :> t)
+
+  exception Interference
+end
 
 let casn_after = Atomic.make `After
 let casn_before = Atomic.make `Before
 
-let rec release_after = function
+let rec release_after casn = function
   | NIL -> true
   | CASN (_, state, lt, gt) ->
-      release_after lt |> ignore;
-      state.before <- state.after;
-      state.casn <- casn_after;
-      release_after gt
+      release_after casn lt |> ignore;
+      if not (is_cmp casn state) then (
+        state.before <- state.after;
+        state.casn <- casn_after);
+      release_after casn gt
 
-let rec release_before = function
+let rec release_before casn = function
   | NIL -> false
   | CASN (_, state, lt, gt) ->
-      release_before lt |> ignore;
-      state.after <- state.before;
-      state.casn <- casn_before;
-      release_before gt
+      release_before casn lt |> ignore;
+      if not (is_cmp casn state) then (
+        state.after <- state.before;
+        state.casn <- casn_before);
+      release_before casn gt
 
-(* Note: The writes to `state.casn <- ...` above could be removed to reduce time
-   at the cost of increasing space usage (by a constant factor). *)
+let release casn cass = function
+  | `After -> release_after casn cass
+  | `Before -> release_before casn cass
 
-let release cass = function
-  | `After -> release_after cass
-  | `Before -> release_before cass
+let rec verify casn = function
+  | NIL -> `After
+  | CASN (atom, desired, lt, gt) -> (
+      match verify casn lt with
+      | `After ->
+          if is_cmp casn desired && Atomic.get atom.state != desired then
+            `Before
+          else verify casn gt
+      | `Before -> `Before)
 
-let finish casn (`Undetermined cass as undetermined)
-    (status : [ `Before | `After ]) =
+let finish casn (`Undetermined cass as undetermined) (status : determined) =
   if Atomic.compare_and_set casn (undetermined :> status) (status :> status)
-  then release cass status
+  then release casn cass status
   else Atomic.get casn == `After
 
-let rec determine casn undetermined skip = function
-  | NIL -> skip || finish casn undetermined `After
-  | CASN (loc, desired, lt, gt) as eq ->
-      determine casn undetermined true lt
-      &&
-      let current = Atomic.get loc.state in
-      if desired == current then determine casn undetermined skip gt
-      else
-        let expected = desired.before in
-        if
-          expected == current.after
-          && (current.casn == casn_after || is_after current.casn)
-          || expected == current.before
-             && (current.casn == casn_before || not (is_after current.casn))
-        then
-          let status = Atomic.get casn in
-          if status != (undetermined :> status) then status == `After
-          else if Atomic.compare_and_set loc.state current desired then
-            determine casn undetermined skip gt
-          else determine casn undetermined skip eq
-        else finish casn undetermined `Before
+let exit _ = raise Exit [@@inline never]
+
+let rec determine casn action = function
+  | NIL -> action
+  | CASN (loc, desired, lt, gt) as eq -> (
+      match determine casn action lt with
+      | `Before -> `Before
+      | (`After | `Verify) as action ->
+          let current = Atomic.get loc.state in
+          if desired == current then
+            let action = if is_cmp casn desired then `Verify else action in
+            determine casn action gt
+          else
+            let matches_expected () =
+              let expected = desired.before in
+              expected == current.after
+              && (current.casn == casn_after || is_after current.casn)
+              || expected == current.before
+                 && (current.casn == casn_before || not (is_after current.casn))
+            in
+            if (not (is_cmp casn desired)) && matches_expected () then
+              match Atomic.get casn with
+              | `Undetermined _ ->
+                  if Atomic.compare_and_set loc.state current desired then
+                    determine casn action gt
+                  else determine casn action eq
+              | #determined -> exit ()
+            else `Before)
 
 and is_after casn =
   match Atomic.get casn with
-  | `Undetermined cass as undetermined -> determine casn undetermined false cass
+  | `Undetermined cass as undetermined -> (
+      match determine casn `After cass with
+      | `Verify -> finish casn undetermined (verify casn cass)
+      | #determined as status -> finish casn undetermined status
+      | exception Exit -> Atomic.get casn == `After)
   | `After -> true
   | `Before -> false
 
@@ -81,7 +113,16 @@ let determine_for_owner casn cass =
   (* The end result is a cyclic data structure, which is why we cannot
      initialize the [casn] atomic directly. *)
   Atomic.set casn undetermined;
-  determine casn undetermined false cass
+  match determine casn `After cass with
+  | `Verify ->
+      (* We only want to [raise Interference] in case it is the verify step that
+         fails.  The idea is that in [lock_free] mode the attempt might have
+         succeeded as the compared locations would have been set in [lock_free]
+         mode preventing interference.  If failure happens before the verify
+         step then the [lock_free] mode would have likely also failed. *)
+      finish casn undetermined (verify casn cass) || raise Mode.Interference
+  | #determined as status -> finish casn undetermined status
+  | exception Exit -> Atomic.get casn == `After
   [@@inline]
 
 let overlap () = failwith "kcas: location overlap" [@@inline never]
@@ -133,6 +174,10 @@ let rec update_no_alloc backoff loc state set_after =
     let backoff = Backoff.once backoff in
     update_no_alloc backoff loc state set_after
 
+let is_obstruction_free casn =
+  Atomic.get casn == (Mode.obstruction_free :> status)
+  [@@inline]
+
 let cas loc before state =
   let state' = Atomic.get loc.state in
   let before' = state'.before and after' = state'.after in
@@ -140,8 +185,6 @@ let cas loc before state =
   || before == if is_after state'.casn then after' else before')
   && Atomic.compare_and_set loc.state state' state
   [@@inline]
-
-let cmp_state casn before = { before; after = before; casn } [@@inline]
 
 module Loc = struct
   type 'a t = 'a loc
@@ -194,58 +237,60 @@ module Op = struct
   type t = CAS : 'a Loc.t * 'a * 'a -> t
 
   let make_cas loc before after = CAS (loc, before, after) [@@inline]
+  let make_cmp loc expected = CAS (loc, expected, expected) [@@inline]
 
   let is_on_loc op loc =
     match op with CAS (loc', _, _) -> Obj.magic loc' == loc
     [@@inline]
 
-  let atomic = function
-    | CAS (loc, before, after) -> Loc.compare_and_set loc before after
+  let get_id = function CAS (loc, _, _) -> loc.id [@@inline]
 
-  let atomically = function
+  let atomic = function
+    | CAS (loc, before, after) ->
+        if before == after then Loc.get loc == before
+        else Loc.compare_and_set loc before after
+
+  let atomically ?(mode = Mode.lock_free) = function
     | [] -> true
     | [ op ] -> atomic op
     | first :: rest ->
-        let casn = Atomic.make `After in
-        let insert cass = function
-          | CAS (loc, before, after) -> insert cass loc { before; after; casn }
+        let casn = Atomic.make (mode :> status) in
+        let rec run cass = function
+          | [] -> determine_for_owner casn cass
+          | CAS (loc, before, after) :: rest ->
+              if before == after && is_obstruction_free casn then
+                let state = Atomic.get loc.state in
+                before == eval state && run (insert cass loc state) rest
+              else run (insert cass loc { before; after; casn }) rest
         in
-        let cass =
-          match first with
-          | CAS (loc, before, after) ->
-              CASN (loc, { before; after; casn }, NIL, NIL)
-        in
-        let cass = List.fold_left insert cass rest in
-        determine_for_owner casn cass
+        let (CAS (loc, before, after)) = first in
+        if before == after && is_obstruction_free casn then
+          let state = Atomic.get loc.state in
+          before == eval state && run (CASN (loc, state, NIL, NIL)) rest
+        else run (CASN (loc, { before; after; casn }, NIL, NIL)) rest
 end
 
-let get_as0 g loc casn l r =
+let update_as0 g loc f casn l r =
   let state = Atomic.get loc.state in
   let before = eval state in
-  let state = cmp_state casn before in
+  let after = f before in
+  let state =
+    if before == after && is_obstruction_free casn then state
+    else { before; after; casn }
+  in
   ((casn, CASN (loc, state, l, r)), g before)
-  [@@inline]
-
-let get_as g loc (casn, cass) =
-  match cass with
-  | NIL -> get_as0 g loc casn NIL NIL
-  | _ -> (
-      match splay loc.id cass with
-      | l, Miss, r -> get_as0 g loc casn l r
-      | l, Hit (_loc', state'), r ->
-          let state = Obj.magic state' in
-          ((casn, CASN (loc, state, l, r)), g state.after))
-  [@@inline]
-
-let update_as0 g loc f casn l r =
-  let before = Loc.get loc in
-  ((casn, CASN (loc, { before; after = f before; casn }, l, r)), g before)
   [@@inline]
 
 let update_as g loc f casn state' l r =
   let state = Obj.magic state' in
-  let current = state.after in
-  ((casn, CASN (loc, { state with after = f current }, l, r)), g current)
+  if is_cmp casn state then
+    let before = eval state in
+    let after = f before in
+    let state = if before == after then state else { before; after; casn } in
+    ((casn, CASN (loc, state, l, r)), g before)
+  else
+    let current = state.after in
+    ((casn, CASN (loc, { state with after = f current }, l, r)), g current)
   [@@inline]
 
 let update_as g loc f (casn, cass) =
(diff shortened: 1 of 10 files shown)


polytypic:
I ran a series of benchmarks to compare the plain GKMZ and the new k-CAS-n-CMP.  As expected, the new algorithm adds a bit of overhead, but also performs significantly better on the parallel CMP benchmark.

```
Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.2 ms    711 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.2 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.1 ms    723 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.1 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 10000' ran
    1.00 ± 0.03 times faster than 'kcas/_build/default/test/benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       8.0 ms ±   0.0 ms    [User: 7.1 ms, System: 0.6 ms]
  Range (min … max):     7.9 ms …   8.6 ms    370 runs
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       8.4 ms ±   0.1 ms    [User: 7.5 ms, System: 0.6 ms]
  Range (min … max):     8.3 ms …   8.8 ms    359 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 2 10000' ran
    1.05 ± 0.01 times faster than 'kcas/_build/default/test/benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      13.3 ms ±   0.1 ms    [User: 12.4 ms, System: 0.6 ms]
  Range (min … max):    13.1 ms …  13.9 ms    226 runs
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      13.9 ms ±   0.1 ms    [User: 13.0 ms, System: 0.6 ms]
  Range (min … max):    13.8 ms …  14.4 ms    215 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 4 10000' ran
    1.05 ± 0.01 times faster than 'kcas/_build/default/test/benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.5 ms    576 runs
 
Benchmark 2: kcas/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.5 ms    571 runs
 
Summary
  'kcas/_build/default/test/tx_benchmark.exe 1 10000' ran
    1.00 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.1 ms ±   0.1 ms    [User: 9.2 ms, System: 0.6 ms]
  Range (min … max):     9.9 ms …  11.1 ms    298 runs
 
Benchmark 2: kcas/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.5 ms ±   0.1 ms    [User: 9.6 ms, System: 0.6 ms]
  Range (min … max):    10.3 ms …  10.9 ms    288 runs
 
Summary
  'kcas-main/_build/default/test/tx_benchmark.exe 2 10000' ran
    1.04 ± 0.01 times faster than 'kcas/_build/default/test/tx_benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      18.5 ms ±   0.2 ms    [User: 17.5 ms, System: 0.7 ms]
  Range (min … max):    18.3 ms …  19.9 ms    161 runs
 
Benchmark 2: kcas/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      18.9 ms ±   0.1 ms    [User: 17.9 ms, System: 0.7 ms]
  Range (min … max):    18.6 ms …  19.3 ms    157 runs
 
Summary
  'kcas-main/_build/default/test/tx_benchmark.exe 4 10000' ran
    1.02 ± 0.01 times faster than 'kcas/_build/default/test/tx_benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      54.5 ms ±   6.9 ms    [User: 95.3 ms, System: 2.4 ms]
  Range (min … max):    41.0 ms …  63.6 ms    71 runs
 
Benchmark 2: kcas/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      21.7 ms ±   1.2 ms    [User: 32.7 ms, System: 2.0 ms]
  Range (min … max):    20.1 ms …  32.3 ms    136 runs
 
Summary
  'kcas/_build/default/test/tx_parallel_cmp_bench.exe 100000' ran
    2.51 ± 0.35 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):     105.0 ms ±  14.0 ms    [User: 186.4 ms, System: 3.7 ms]
  Range (min … max):    80.4 ms … 120.1 ms    25 runs
 
Benchmark 2: kcas/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):      40.8 ms ±   1.7 ms    [User: 63.4 ms, System: 3.0 ms]
  Range (min … max):    38.3 ms …  45.1 ms    70 runs
 
Summary
  'kcas/_build/default/test/tx_parallel_cmp_bench.exe 200000' ran
    2.58 ± 0.36 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):     213.0 ms ±  25.4 ms    [User: 379.9 ms, System: 6.5 ms]
  Range (min … max):   158.8 ms … 238.0 ms    12 runs
 
Benchmark 2: kcas/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):      80.3 ms ±   4.6 ms    [User: 126.8 ms, System: 5.0 ms]
  Range (min … max):    74.5 ms …  98.6 ms    36 runs
 
Summary
  'kcas/_build/default/test/tx_parallel_cmp_bench.exe 400000' ran
    2.65 ± 0.35 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      34.4 ms ±   0.6 ms    [User: 33.4 ms, System: 0.7 ms]
  Range (min … max):    33.5 ms …  37.7 ms    87 runs
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      34.9 ms ±   0.3 ms    [User: 33.9 ms, System: 0.7 ms]
  Range (min … max):    33.9 ms …  35.6 ms    86 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 200000' ran
    1.01 ± 0.02 times faster than 'kcas/_build/default/test/benchmark.exe 1 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     127.6 ms ±   0.3 ms    [User: 126.2 ms, System: 1.0 ms]
  Range (min … max):   127.1 ms … 128.1 ms    23 runs
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     134.8 ms ±   0.3 ms    [User: 133.5 ms, System: 1.0 ms]
  Range (min … max):   134.4 ms … 135.4 ms    22 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 2 200000' ran
    1.06 ± 0.00 times faster than 'kcas/_build/default/test/benchmark.exe 2 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     233.7 ms ±   0.3 ms    [User: 232.1 ms, System: 1.2 ms]
  Range (min … max):   233.4 ms … 234.4 ms    12 runs
 
Benchmark 2: kcas/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     246.0 ms ±   0.4 ms    [User: 244.3 ms, System: 1.3 ms]
  Range (min … max):   245.4 ms … 246.6 ms    12 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 4 200000' ran
    1.05 ± 0.00 times faster than 'kcas/_build/default/test/benchmark.exe 4 200000'
```

src/kcas.ml
```diff
@@ -194,35 +238,43 @@ module Op = struct
...
-          | CAS (loc, before, after) ->
-              CASN (loc, { before; after; casn }, NIL, NIL)
-        in
-        let cass = List.fold_left insert cass rest in
-        determine_for_owner casn cass
+        let (CAS (loc, before, after)) = first in
+        if before == after && is_obstruction_free casn then
+          let state = Atomic.get loc.state in
+          before == eval state && run (CASN (loc, state, NIL, NIL)) rest
+        else run (CASN (loc, { before; after; casn }, NIL, NIL)) rest
```
bartoszmodelski, line 269: Can the duplication above (250-263 and 266-269) be avoided?

src/kcas.ml
```diff
@@ -194,35 +238,43 @@ module Op = struct
...
     | [ op ] -> atomic op
     | first :: rest ->
-        let casn = Atomic.make `After in
-        let insert cass = function
-          | CAS (loc, before, after) -> insert cass loc { before; after; casn }
+        let casn = Atomic.make (mode :> status) in
+        let rec run cass = function
+          | [] -> determine_for_owner casn cass
+          | CAS (loc, before, after) :: rest ->
+              if before == after && is_obstruction_free casn then
```
bartoszmodelski, line 260: Mode is given as argument but it's re-read from atomic var here. Can mode actually change mid-way through execution?

src/kcas.ml
```diff
@@ -194,35 +238,43 @@ module Op = struct
   type t = CAS : 'a Loc.t * 'a * 'a -> t
 
   let make_cas loc before after = CAS (loc, before, after) [@@inline]
+  let make_cmp loc expected = CAS (loc, expected, expected) [@@inline]
```
bartoszmodelski, line 241: There's a few 3-letter abbreviations in the space already. Perhaps `make_compare`?

src/kcas.ml
```diff
@@ -194,35 +238,43 @@ module Op = struct
...
 
   let is_on_loc op loc =
     match op with CAS (loc', _, _) -> Obj.magic loc' == loc
```
bartoszmodelski, line 244: Not added here, so feel free to punt this. We could just compare loc's id's here, rather than use magic, right?

src/kcas.ml
```diff
@@ -10,69 +10,102 @@ module Id = struct
...
-          else determine casn undetermined skip eq
-        else finish casn undetermined `Before
+let exit _ = raise Exit [@@inline never]
+
+let rec determine casn action = function
+  | NIL -> action
+  | CASN (loc, desired, lt, gt) as eq -> (
+      match determine casn action lt with
+      | `Before -> `Before
+      | action ->
```
bartoszmodelski, line 79: I'd vote against using wildcards with polymorphic variants. They can be broken with a typo, and the code will still compile. Personally, I try to avoid them in general, just because it's easier to evolve such a code base. 

```suggestion
      | (`After | `Verify) as action ->
```

src/kcas.ml
```diff
@@ -10,69 +10,102 @@ module Id = struct
...
+            determine casn action gt
+          else
+            let expected = desired.before in
+            if
+              (not (is_cmp casn desired))
+              && (expected == current.after
+                  && (current.casn == casn_after || is_after current.casn)
+                 || expected == current.before
+                    && (current.casn == casn_before
+                       || not (is_after current.casn)))
```
bartoszmodelski, line 92: This condition is getting complicated; perhaps it can be divided into components with meaningful names?

src/kcas.ml
```diff
@@ -10,69 +10,102 @@ module Id = struct
...
+
+let rec verify casn = function
+  | NIL -> `After
+  | CASN (atom, desired, lt, gt) -> (
+      match verify casn lt with
+      | `After ->
+          if is_cmp casn desired && Atomic.get atom.state != desired then
+            `Before
+          else verify casn gt
+      | status -> status)
```
bartoszmodelski, line 65: See comment on line 79.

src/kcas.mli
```diff
@@ -87,15 +112,19 @@ module Tx : sig
...
       2. The second phase attempts to perform the operations atomically.
 
-      Either phase may fail.  In particular, the first phase is allowed to raise
-      exceptions to signal failure.
+      3. In {!Mode.obstruction_free} a third phase verifies all read-only
+      operations.
+
+      Each phase may fail.  In particular, the first phase is allowed to raise
+      exceptions to signal failure.  Failure on the third phase raises
+      {!Mode.Interference}.
```
bartoszmodelski, line 127: This adds complexity to the otherwise pretty simple `atomically`. Is there a good reason to not just return false on interference?

src/kcas.ml
```diff
@@ -259,6 +324,29 @@ let update_as g loc f (casn, cass) =
       | l, Hit (_loc', state'), r -> update_as g loc f casn state' l r)
   [@@inline]
 
+let attempt (mode : Mode.t) xt =
```
bartoszmodelski, line 327: `tx` I suspect?

src/kcas.ml
```diff
@@ -10,69 +10,102 @@ module Id = struct
...
+                 || expected == current.before
+                    && (current.casn == casn_before
+                       || not (is_after current.casn)))
+            then
+              match Atomic.get casn with
+              | `Undetermined _ ->
+                  if Atomic.compare_and_set loc.state current desired then
+                    determine casn action gt
+                  else determine casn action eq
+              | #determined -> exit ()
```
bartoszmodelski, line 99: Why raise rather than return observed outcome?

bartoszmodelski (commented):
Thanks for this PR! 

As I said, it's going to take me some time to review this. I've added the first tranche of comments; feel free to wait until I'm done or handle them as they flow in.

src/kcas.ml
polytypic: Hmm... I don't have a strong opinion on this, but I kind of like having it the same length with `cas`. 😄

I'm Ok with either `make_cmp` or `make_compare`.

Other votes?

src/kcas.ml
polytypic: No, it cannot change.

The `casn` atomic location is otherwise unused until the full `cass` splay tree is ready and then the `casn` is set to `Undetermined cass` in `determine_for_owner` before calling `determine`.

I pass the mode through the `casn` in the transaction mechanism, because that way every `Tx` operation will potentially allocate one word less:

```diff
-type log = casn * cass
+type log = mode * casn * cass
```

I then use the same piggybacking in `atomically` for consistency.

Admittedly it is a bit of hack/might not be immediately obvious, but I tried a number of approaches and this seemed to give low overhead and `is_obstruction_free casn` accurately reads as "is this CASN going to be attempted in obstruction free mode".

src/kcas.mli
polytypic: Yes, I believe there is.

See the comment in `determine_for_owner`.  It is useful to know whether failure was detected during the post hoc verification step.  It allows the caller to make a better decision on whether it makes sense to continue trying in obstruction-free mode or to switch to lock-free mode.  If failure happens before post hoc verification, switching to lock-free mode would likely not (have) help(ed).

My assumption is that `atomically` will only be used rarely by experts.  I intentionally chose the default for `atomically` and `attempt` to be `lock_free`.  The `Interference` exception will not be raised unless `obstruction_free` mode is explicitly requested.

On the other hand, `commit` takes care of handling the `Interference` exception internally.  I would assume most users will be using the `Tx` interface with `commit` and would also recommend that as `commit` also implements the retry loop and backoff (doing those correctly can be tricky enough).  Perhaps this should be made more explicit in the documentation (i.e. that `attempt` and the `Op` interface are not intended for casual users).

Also, considering the potential support for blocking transactions, there one would write the (blocking) equivalent of `commit` outside of kcas and then it would be ideal for such an implementation to be able to make the best decisions on when to switch to lock-free mode.  So, having `attempt` also raise `Interference` in case the post hoc verification failed.

src/kcas.ml
polytypic: Yeah...  It wasn't a typo, but something I failed to rename at some point. 😅

I noticed that already earlier, but didn't bother to change it.  I'll change those.

src/kcas.ml
polytypic: The reason for this is to keep the conditional that determines whether to continue the tree traversal essentially as a single comparison:

```ocaml
      match determine casn action lt with
      | `Before -> `Before
      | action ->
```

This is one of the spots in the implementation that took me numerous iterations.  The goal is to avoid extra overhead from the choice of two modes and the approach in the PR is the best (lowest overhead) I could find.

The previous algorithm, which only had a lock-free mode, is simpler as it only returns a boolean outcome.  The new version cannot be implemented so simply, because there are more outcomes.

In the new algorithm, the result of `determine` is either `Before` (potential failure), `After` (potential success and no verification required), `Verify` (potential success requiring verification of compare operations), or `exception Exit` indicating that the `casn` was already determined by some other domain.  Separating all of these cases efficiently is necessary for best performance.

src/kcas.ml
polytypic: Note that the `action` value is the return value of the `determine` function and that function is only called from `is_after` and `determine_for_owner`, which both explicitly handle all the cases.  Also, the polymorphic variants ultimately flow into the monotyped `casn` atomic.  So, basically, if you make typo it will be caught.

But it shouldn't hurt to be more explicit in the couple of places you mentioned, so I'll change those spots to be more explicit.

src/kcas.ml
polytypic: I added `Op.get_id` to the PR.

src/kcas.ml
polytypic: I agree, the conditional here is hairy.  The annoying problem here (again) is that that the existential in the GADT prevents some refactorings (or at least makes the harder/more expensive (I mean, we could avoid all the magic using Leibniz equality, but I wouldn't go there)), but I moved the original conditional to a local helper function and perhaps the main intent is now clearer.

src/kcas.ml
polytypic: Hmm... The duplication here is a bit unfortunate, but it is difficult to eliminate without adding overhead.  Recursion tends to limit compiler optimizations and there is also nested conditionals here (`if` and also `&& run`).  It is basically an explicit manual unrolling of the logic for the `first` logical CAS and it does give measurable overhead reduction.  It also plays together with the optimization in `insert` that explicitly checks for the left/right leaning trees.  These optimizations (unroll the loop once and special case for presorted) were already introduced in the previous implementation.

BTW, unrelated to this.  I realized that `Tx.get` and `Tx.get_as` can be expressed in terms of `Tx.update_as` and in those cases it doesn't seem to be more expensive, which is probably partly due to those not being recursive functions, so the compiler will be able to optimize them after inlining.  I just removed the duplication and now `Tx.get` and `Tx.get_as` are implemented in terms of the internal `update_as` (all the location accesses now are).

src/kcas.mli
```diff
@@ -45,6 +45,25 @@ module Loc : sig
...
+  (** Type of an operating mode of the [k-CAS-n-CMP] algorithm. *)
+
+  val lock_free : t
+  (** In [lock_free] mode the algorithm makes sure that at least one domain will
+      be able to make progress. *)
+
+  val obstruction_free : t
+  (** In [obstruction_free] mode the algorithm proceeds optimistically and
+      allows operations to fail due to {!Interference} from other domains that
+      might have been prevented in the {!lock_free} mode. *)
```
bartoszmodelski, line 60: We should highlight that interference can look spurious, i.e. value of some loc does not have to change for interference to be raised.

bartoszmodelski:
I spent a while looking at this and I'm convinced the approach here is sound. I left another comment but other than that, we should be good to go. Clever stuff!

src/kcas.mli
bartoszmodelski: The corollary here would be that some locations are best compared in LF mode, while others in OF and a potentially more useful API would let user specify that, rather than force all compares into one mode of execution. This naturally doesn't have to come in this PR.

polytypic: > is there a good use-case for retrying in lock-free mode after failure in obstruction-mode?

I guess what I have in mind with the approach of starting, by default, in OF mode and then switching to LF mode is that it could ideally just work/perform well enough that most users would not need to think about it.

Why should it work well?  Well, if you have a relatively low contention scenario, then OF mode is likely to succeed.  So, it should make sense to go with OF mode as it would (reduce the/maintain the low) amount of writes and contention.

In a high contention scenario, OTOH, it might make sense to start in LF mode, because OF would be relatively more likely to fail and staying in OF might even cause starvation.  However, switching quickly to LF mode on interference should also mean that not much is lost by trying in OF mode first even with higher contention.  That is at least how I would intuitively assume.  It could be interesting to try to model this mathematically or through simulations.

Note that one can already explicitly start `commit` in lock-free mode and it is also possible to implement `commit` on top of `attempt` and choose desired strategy for dealing with failure and interference.

> some locations are best compared in LF mode

Hmm... So, you mean that the mode to use would be specified when creating the _location_ (rather then when calling `commit`)?  That is an interesting idea, indeed!

Hmm... Perhaps something could even be done automatically.  Perhaps something like branch-prediction.  For example, there could be some sort of memory maintained on failed compare verifies with each location.  To keep things low-overhead, that memory would only be updated in case of verify failure, but it could be consulted e.g. when creating the `CASN` descriptors (in the internal `update_as`, for example).  If the location seems "hot" (lots of compare verify failures), then the algorithm would just create a `CAS` against that location.

I think there could be interesting work here to do once we have more examples and benchmarks using this.

src/kcas.mli
polytypic: I added a commit with a bit of documentation on this.

bartoszmodelski (approved):

polytypic:
> I spent a while looking at this and I'm convinced the approach here is sound. I left another comment but other than that, we should be good to go. Clever stuff!

Thanks for the thorough review!

PR #26, opened 2023-02-10
Restructure README and add introduction with mdx tests

polytypic:
This PR restructures the README and adds an introduction to the library with mdx checked examples.

(Note that the benchmarks failing is not related to this PR.)


README.md
```diff
@@ -1,35 +1,455 @@
...
+```ocaml
+# let rec move ?(backoff = Backoff.default)
+               source
+               target =
+    match Loc.get source with
+    | [] -> raise Exit
+    | (elem::rest) as source' ->
+      let target' = Loc.get target in
+      let ops = [
+        Op.make_cas source source' rest;
```
talex5, line 175: The use of `source'` to mean "the old value of `source`" is the opposite to other uses I've seen of `'`, which might be confusing.

polytypic: Hmm... How about `source_loc` and `target_loc` for the parameters?

polytypic: Or perhaps `old_source` and `old_target`, instead.  I guess that is clearer.

talex5 (approved):
Looks good to me!

PR #27, opened 2023-02-13
Benchmarks CI integration

bartoszmodelski:
Adds integration with [current-bench](https://github.com/ocurrent/current-bench) in advance of the meeting with benchmarking team. 

The CI is already running on this project (and failing). This change implements `make bench`, which executes all benchmarks of interest and prints adequately formatted results. 

* The first commit contains only single-core benchmarks and these should succeed. 
* Once the single-core benchmarks pass, I will push multicore ones. I expect these to fail due to issue with benchmarking infra. 

Update: holding off with the multicore benchmarks since it seems that the single-core ones will require some troubleshooting, as data does not show [here](https://autumn.ocamllabs.io/ocaml-multicore/kcas/pull/27?worker=autumn&image=ocaml%2Fopam%3Adebian-ocaml-5.0) despite the success.


bartoszmodelski:
This now works: https://autumn.ocamllabs.io/ocaml-multicore/kcas/pull/27/benchmark/kcas?worker=autumn&image=ocaml%2Fopam%3Adebian-ocaml-5.0

polytypic:
How close is this to being ready?  I'd love to see the benchmarks running...

bartoszmodelski:
Everything on our side is ready. On the benchmarking side, they are able to run only the single-core benchmarks reliably but they know the issues and are working towards the resolution. 

So, I'd suggest going forward with this PR (given the CI succeeds), as single-core benchmarks already give some signal, and once we're ready we can just move over the parallel ones.

bartoszmodelski:
Perhaps two things for me to add in this PR if we want to move forward with single-core:  
* For consistency and to make the future switch as simple as possible, I can move the parallel tests under the same "orchestration" and just have them run on `make bench-parallel` rather than `make bench`.
* Add a note in the readme.

polytypic:
Thanks for starting this work!  I ended up adding ocurrent-bench integration via another PR #94.

PR #30, opened 2023-02-18
Add xplicit transaction passing API

polytypic:
Thanks to [suggestion by Gabriel Scherer](https://discuss.ocaml.org/t/ann-major-updates-to-kcas-in-0-1-8-and-0-2-0/11392/3)!

This PR adds an API for recording a transaction log by explicitly passing around a mutable transaction log.  The main advantage of this approach is that it avoids the use of (monadic) combinators and allows convenient use of all ordinary sequential control structures of OCaml.

To make this approach safe(r), the implementation uses the same trick as in the classic paper [Lazy Functional State Threads](https://dl.acm.org/doi/10.1145/178243.178246).  Basically, the mutable transaction log type has a phantom type variable and the only way to perform a log is through functions that must be universally quantified over the transaction log.  This means that attempts to leak and reuse a log are prevented by the type system.  Of course, it is still possible to capture the log inside a closure, for example.  This can cause no harm, however, unless such a captured log is used concurrently in another domain, because the transaction log construction is not domain-safe itself. While possible, such misuse requires effort and should be unlikely to happen accidentally.

This PR restructures the documentation to document the three-phase protocol separately as it applies to all the three different APIs to perform operations over multiple shared memory locations atomically.  This PR also adds some structure to the `Tx` module documentation and some other touches that are unrelated to the changes in this PR.

This PR also adds a variation of the parallel compare benchmark using the added`Xt` API.  Performance seems comparable to the previous `Tx` based version of that benchmark.  More benchmarking would be needed! The `Xt` API likely imposes more GC write-barriers while the `Tx` API potentially incurs more memory allocations.

This PR also rewrites the 2-stack queue into a 3-stack queue using the `Xt` API and then the tests of `Xt` and `Tx` use the conversions provided by the `Xt` API to test and demonstrate the interoperability between the two APIs.

This PR adds documentation on the new `Xt` API to the main README.  The new example is based on an imperative leftist heap implementation.  The idea being to make the examples progressively more complicated.

This PR also implements the same presorted access optimization for the `Tx` (and `Xt`) API as was previously implemented for the `Op` API.


gasche:
There is a lot of code duplication with slightly-different variants of the same functions. This could be avoid by turning `update_as` to CPS style (instead of `... -> casn * cass * 'a`, make that `(casn -> cass -> 'a -> 'r) -> 'r`), and then using it in `Tx` to return `log * 'a` and in `Xt` to mutate the input and return just `'a`.

polytypic:
> There is a lot of code duplication with slightly-different variants of the same functions. This could be avoid by turning `update_as` to CPS style (instead of `... -> casn * cass * 'a`, make that `(casn -> cass -> 'a -> 'r) -> 'r`), and then using it in `Tx` to return `log * 'a` and in `Xt` to mutate the input and return just `'a`.

Yes, CPS can reduce duplication here.

I made a quick experiment to use CPS and unfortunately it seems to perform worse.  I assume the compiler is not quite as capable inlining CPS style functions.  The amount of duplication doesn't feel high enough to warrant reducing performance by 5%-20% (based on benchmarks against `main`).  But this is certainly worth revisiting if there is some specific way to coerce the compiler to perform the necessary optimizations to avoid performance regression.

Note that the implementation already intentionally has some amount of duplication (e.g. perform first iteration of a loop outside the loop) for performance reasons.

polytypic:
In a private conversation, @kayceesrk asked about the possibility of breaking the abstraction by spawning a domain.  That is a good question, because it is well known that multi-shot continuations basically allow one to break all sorts of higher-order mutable abstractions.

I believe the implementation can be considered reasonably safe against that.  There are a couple of reasons for that.

First of all, a log can only be `attempt`ed or `commit`ed through a transaction function that needs to be polymorphic with respect to the log.  So, even if a domain could be spawned inside that transaction function and retain reference to the log, there is no way to `attempt` or `commit` that log (without using unsafe features of OCaml).

The log itself, however, contains both atomic and non-atomic mutable locations and it is not thread/domain safe &mdash; log construction needs to happen within a single thread of execution and we should document that.  However, the only location that is being mutated during the log construction is the `mutable cass : cass` field.  This means that even if the log would be captured by a domain and then modified after the log has been `attempt`ed, such mutations could not change any of the k-CAS locations.  However, it is possible to (un)safely mutate the `mutable cass : cass` field and change the log before the log `attempt` is finished, just like with ordinary captured `ref`s or other mutable fields.

gasche:
My understanding of your point is as follows: if we move from a persistent API (returning a new state) to a mutable API (mutating the previous state), do we get in trouble on unsynchronized concurrent usage? In other words, now we have to decide whether the log value is domain-safe or not -- if we wanted to make it domain-safe we would need to use `Atomic.t` which incurs extra overhead. You seem intent on documenting this as "not domain-safe", and this seems fine.


More prose on this, less important:
- The implementation choices may influence what users observe if they use the data-structure incorrectly. Correctly you use a record with two fields that are mutated independently. You could also use a mutable reference to a pair of states (that is, exactly `log ref` with your previous `log` type); then data races would never break the guarantee that the two elements of the pair have been produced together, they represent a valid log. You would, at worst, only miss some of the log operations performed by other domains.
- You could also equip the structure with an atomic version counter to detect incorrect usage and fail early. But then again, this is probably overkill for something that is only going to be used in short sections of sequential code performing lock-free operations.

polytypic:
> You seem intent on documenting this as "not domain-safe", and this seems fine.

Yes.

> You would, at worst, only miss some of the log operations performed by other domains.

Hmm... I think it is worse than that.  Log update in (another) domain spawned inside a transaction could read `cass` at the start (when it contains `NIL`) and then update the `cass` just before the initiator domain reads `cass`. All the operations made by the initiator domain would be lost and the `result` computed by the initiator domain would be completely inconsistent with the log.

Whether using a mutable field (i.e. `casn * mutable cass`) or reference (i.e. `(casn * cass) ref`) doesn't change that because the `casn` part of the log is (for the updates of the `cass` reference) constant so there is no difference in level of consistency between the two implementations.  I wondered about this myself, but upon a closer look, there is no difference.

But, yes, I would say it is fine to say that the log itself is not domain-safe and adding safety guards to it seems overkill to me at this point, because misuse basically requires cleverness (you need to spawn something that lives beyond the scope) and seems reasonably unlikely to happen accidentally.

I think that to make the log domain-safe, one could change the log updates to be performed atomically.  IOW, instead of just writing to `xt.cass`

```ocaml
    xt.cass <- CASN (loc, state, l, r);
```

the updates would be done using `compare_and_set`

```ocaml
  let old_cass = Atomic.get xt.cass
  ...
  if not (Atomic.compare_and_set xt.cass old_cass new_cass) then
    retry_log_update ...
  else
    ...
```

This should work to ensure that updates to the log are atomic, because the `cass` updates always allocate so the value of `cass` is always unique after each update.

The problem with this is that it would likely make the log updates substantially slower, because then they would be done inside a retry loop (inhibiting compiler optimizations) and using slower atomic operations.

gasche:
(A very very minor point: due to my at-name in your commit message, I get two notification emails each time you rebase this PR, so I got about 80 right now. It's a silly aspect of the Github interface, but for this reason I generally prefer to use full names or just nicknames without the at-mention sign in commit messages; when the information is available I use the semi-standard `Suggested-by: name <email>` form. Thanks for the explicit crediting anyway, it's a nice touch.)

polytypic:
> so I got about 80 right now

Ouch! Sorry about that!

I didn't realise GH sends so many emails as I personally tend to always disable email notifications from every service I use (and I have disabled those on GH also).

BTW, I think the PR is pretty much ready.  I just made some tweaks to the documentation.  I'll update the main note with some more information on the PR shortly.

src/kcas.ml
```diff
@@ -294,12 +294,17 @@ let update_as g loc f casn state' l r =
   [@@inline]
 
 let update_as g loc f (casn, cass) =
+  let x = loc.id in
   match cass with
   | NIL -> update_as0 g loc f casn NIL NIL
+  | CASN (a, _, NIL, _) as cass when x < a.id ->
+      update_as0 g loc f casn NIL cass
+  | CASN (a, _, _, NIL) as cass when a.id < x ->
+      update_as0 g loc f casn cass NIL
```
polytypic, line 303: These implement the same presorted access optimization as is done with the `Op` API.

polytypic:
```
Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.1 ms    730 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.1 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.1 ms    731 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.1 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 10000' ran
    1.00 ± 0.02 times faster than 'kcas-this/_build/default/test/benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       8.1 ms ±   0.1 ms    [User: 7.2 ms, System: 0.6 ms]
  Range (min … max):     8.0 ms …   8.5 ms    368 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       7.8 ms ±   0.1 ms    [User: 6.9 ms, System: 0.6 ms]
  Range (min … max):     7.7 ms …   8.2 ms    387 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 2 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      13.4 ms ±   0.1 ms    [User: 12.5 ms, System: 0.6 ms]
  Range (min … max):    13.2 ms …  13.8 ms    224 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      12.9 ms ±   0.1 ms    [User: 12.0 ms, System: 0.6 ms]
  Range (min … max):    12.7 ms …  13.3 ms    231 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 4 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.6 ms    573 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.7 ms    575 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 1 10000' ran
    1.00 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.5 ms ±   0.1 ms    [User: 9.6 ms, System: 0.6 ms]
  Range (min … max):    10.4 ms …  10.9 ms    283 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.1 ms ±   0.1 ms    [User: 9.2 ms, System: 0.7 ms]
  Range (min … max):    10.0 ms …  10.9 ms    295 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 2 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      18.9 ms ±   0.1 ms    [User: 17.9 ms, System: 0.7 ms]
  Range (min … max):    18.8 ms …  20.1 ms    159 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      17.9 ms ±   0.2 ms    [User: 16.9 ms, System: 0.7 ms]
  Range (min … max):    17.7 ms …  19.1 ms    167 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 4 10000' ran
    1.06 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      21.8 ms ±   0.8 ms    [User: 32.8 ms, System: 2.0 ms]
  Range (min … max):    20.4 ms …  23.5 ms    144 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      20.2 ms ±   1.0 ms    [User: 30.4 ms, System: 1.9 ms]
  Range (min … max):    18.7 ms …  21.9 ms    142 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 100000' ran
    1.08 ± 0.07 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):      40.8 ms ±   1.8 ms    [User: 63.8 ms, System: 2.9 ms]
  Range (min … max):    38.6 ms …  47.0 ms    69 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):      38.5 ms ±   2.0 ms    [User: 60.6 ms, System: 2.8 ms]
  Range (min … max):    35.1 ms …  42.6 ms    83 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 200000' ran
    1.06 ± 0.07 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):      79.7 ms ±   5.1 ms    [User: 126.2 ms, System: 4.9 ms]
  Range (min … max):    75.3 ms … 102.9 ms    36 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):      75.0 ms ±   3.8 ms    [User: 120.5 ms, System: 4.6 ms]
  Range (min … max):    68.3 ms …  78.5 ms    43 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 400000' ran
    1.06 ± 0.09 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      35.2 ms ±   0.3 ms    [User: 34.2 ms, System: 0.7 ms]
  Range (min … max):    34.6 ms …  35.7 ms    86 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      35.3 ms ±   0.3 ms    [User: 34.2 ms, System: 0.7 ms]
  Range (min … max):    34.6 ms …  35.7 ms    86 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 200000' ran
    1.00 ± 0.01 times faster than 'kcas-this/_build/default/test/benchmark.exe 1 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     129.4 ms ±   0.4 ms    [User: 128.0 ms, System: 1.0 ms]
  Range (min … max):   128.6 ms … 130.0 ms    23 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     121.7 ms ±   0.4 ms    [User: 120.3 ms, System: 1.0 ms]
  Range (min … max):   121.1 ms … 122.4 ms    24 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 2 200000' ran
    1.06 ± 0.00 times faster than 'kcas-main/_build/default/test/benchmark.exe 2 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     235.0 ms ±   0.3 ms    [User: 233.4 ms, System: 1.2 ms]
  Range (min … max):   234.5 ms … 235.5 ms    12 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     224.6 ms ±   0.4 ms    [User: 223.0 ms, System: 1.2 ms]
  Range (min … max):   224.1 ms … 225.5 ms    13 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 4 200000' ran
    1.05 ± 0.00 times faster than 'kcas-main/_build/default/test/benchmark.exe 4 200000'
```

polytypic:
```
Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.1 ms    730 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.1 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 1 10000
  Range (min … max):     3.4 ms …   4.1 ms    731 runs
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (4.1 ms). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 10000' ran
    1.00 ± 0.02 times faster than 'kcas-this/_build/default/test/benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       8.1 ms ±   0.1 ms    [User: 7.2 ms, System: 0.6 ms]
  Range (min … max):     8.0 ms …   8.5 ms    368 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 2 10000
  Time (mean ± σ):       7.8 ms ±   0.1 ms    [User: 6.9 ms, System: 0.6 ms]
  Range (min … max):     7.7 ms …   8.2 ms    387 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 2 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      13.4 ms ±   0.1 ms    [User: 12.5 ms, System: 0.6 ms]
  Range (min … max):    13.2 ms …  13.8 ms    224 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 4 10000
  Time (mean ± σ):      12.9 ms ±   0.1 ms    [User: 12.0 ms, System: 0.6 ms]
  Range (min … max):    12.7 ms …  13.3 ms    231 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 4 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.6 ms    573 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 1 10000
  Time (mean ± σ):       5.2 ms ±   0.0 ms    [User: 4.3 ms, System: 0.6 ms]
  Range (min … max):     5.1 ms …   5.7 ms    575 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 1 10000' ran
    1.00 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 1 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.5 ms ±   0.1 ms    [User: 9.6 ms, System: 0.6 ms]
  Range (min … max):    10.4 ms …  10.9 ms    283 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 2 10000
  Time (mean ± σ):      10.1 ms ±   0.1 ms    [User: 9.2 ms, System: 0.7 ms]
  Range (min … max):    10.0 ms …  10.9 ms    295 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 2 10000' ran
    1.04 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 2 10000'


Benchmark 1: kcas-main/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      18.9 ms ±   0.1 ms    [User: 17.9 ms, System: 0.7 ms]
  Range (min … max):    18.8 ms …  20.1 ms    159 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_benchmark.exe 4 10000
  Time (mean ± σ):      17.9 ms ±   0.2 ms    [User: 16.9 ms, System: 0.7 ms]
  Range (min … max):    17.7 ms …  19.1 ms    167 runs
 
Summary
  'kcas-this/_build/default/test/tx_benchmark.exe 4 10000' ran
    1.06 ± 0.01 times faster than 'kcas-main/_build/default/test/tx_benchmark.exe 4 10000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      21.8 ms ±   0.8 ms    [User: 32.8 ms, System: 2.0 ms]
  Range (min … max):    20.4 ms …  23.5 ms    144 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 100000
  Time (mean ± σ):      20.2 ms ±   1.0 ms    [User: 30.4 ms, System: 1.9 ms]
  Range (min … max):    18.7 ms …  21.9 ms    142 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 100000' ran
    1.08 ± 0.07 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 100000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):      40.8 ms ±   1.8 ms    [User: 63.8 ms, System: 2.9 ms]
  Range (min … max):    38.6 ms …  47.0 ms    69 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 200000
  Time (mean ± σ):      38.5 ms ±   2.0 ms    [User: 60.6 ms, System: 2.8 ms]
  Range (min … max):    35.1 ms …  42.6 ms    83 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 200000' ran
    1.06 ± 0.07 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 200000'


Benchmark 1: kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):      79.7 ms ±   5.1 ms    [User: 126.2 ms, System: 4.9 ms]
  Range (min … max):    75.3 ms … 102.9 ms    36 runs
 
Benchmark 2: kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 400000
  Time (mean ± σ):      75.0 ms ±   3.8 ms    [User: 120.5 ms, System: 4.6 ms]
  Range (min … max):    68.3 ms …  78.5 ms    43 runs
 
Summary
  'kcas-this/_build/default/test/tx_parallel_cmp_bench.exe 400000' ran
    1.06 ± 0.09 times faster than 'kcas-main/_build/default/test/tx_parallel_cmp_bench.exe 400000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      35.2 ms ±   0.3 ms    [User: 34.2 ms, System: 0.7 ms]
  Range (min … max):    34.6 ms …  35.7 ms    86 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 1 200000
  Time (mean ± σ):      35.3 ms ±   0.3 ms    [User: 34.2 ms, System: 0.7 ms]
  Range (min … max):    34.6 ms …  35.7 ms    86 runs
 
Summary
  'kcas-main/_build/default/test/benchmark.exe 1 200000' ran
    1.00 ± 0.01 times faster than 'kcas-this/_build/default/test/benchmark.exe 1 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     129.4 ms ±   0.4 ms    [User: 128.0 ms, System: 1.0 ms]
  Range (min … max):   128.6 ms … 130.0 ms    23 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 2 200000
  Time (mean ± σ):     121.7 ms ±   0.4 ms    [User: 120.3 ms, System: 1.0 ms]
  Range (min … max):   121.1 ms … 122.4 ms    24 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 2 200000' ran
    1.06 ± 0.00 times faster than 'kcas-main/_build/default/test/benchmark.exe 2 200000'


Benchmark 1: kcas-main/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     235.0 ms ±   0.3 ms    [User: 233.4 ms, System: 1.2 ms]
  Range (min … max):   234.5 ms … 235.5 ms    12 runs
 
Benchmark 2: kcas-this/_build/default/test/benchmark.exe 4 200000
  Time (mean ± σ):     224.6 ms ±   0.4 ms    [User: 223.0 ms, System: 1.2 ms]
  Range (min … max):   224.1 ms … 225.5 ms    13 runs
 
Summary
  'kcas-this/_build/default/test/benchmark.exe 4 200000' ran
    1.05 ± 0.00 times faster than 'kcas-main/_build/default/test/benchmark.exe 4 200000'
```

src/kcas.mli
```diff
@@ -323,3 +349,85 @@ module Tx : sig
...
+
+  val incr : xt:'x t -> int Loc.t -> unit
+  (** [incr ~xt r] is equivalent to [fetch_and_add ~xt r 1 |> ignore]. *)
+
+  val decr : xt:'x t -> int Loc.t -> unit
+  (** [decr ~xt r] is equivalent to [fetch_and_add ~xt r (-1) |> ignore]. *)
+
+  (** {1 Performing accesses} *)
+
+  type 'a tx = { tx : 'x. xt:'x t -> 'a }
```
samoht, line 398: What's the advantage of using a label here? What's wrong with:

```ocaml
type 'a tx = { tx : 'x. x t -> 'a }
```

The exposed API (a record) is already a bit heavy - maybe there's value in not adding that extra label as well? There's perhaps a clever way to hide this behind a constructor. Another thing to consider here: exposing concrete types often leads to headaches for future API evolutions.

samoht (approved):

src/kcas.mli
polytypic: The label helps to make the explicit log passing stand out (which I consider a good thing) and perhaps make it more idiomatic.  On the other hand, it also allows the label to be omitted (regardless of its position) as in a call like `Xt.commit { tx = Queue.pop_front_opt queue }`.

The reason for having the record is to require the function be polymorphic with respect to the log.  IOW, it is using rank-2 polymorphism.  This makes it difficult to accidentally leak or reuse the log (the `~xt` parameter), which would be unsafe.  IOW, an attempt to e.g. store the value `~xt` to a ref cell allocated outside the transaction and then reuse it in another transaction will be prevented by the type system.

samoht: Ok, in that case, I suggest that the examples could use the idiomatic API.

i.e. in the example, always define an external function with an extra non-labelled parameter and use it. This way, it's extra-clear why you ask users to write such convoluted code :-)

polytypic:
@lyrm Noticed an issue with the `Xt.update` and `Xt.modify` operations.  See comments [here](https://github.com/ocaml-multicore/kcas/pull/33#discussion_r1130838536) and [here](https://github.com/ocaml-multicore/kcas/pull/33#discussion_r1130843309).

Thanks!  That is an excellent catch!

The problem is that the internal higher-order `update` assumes that the function passed does not mutate the transaction log.

I see a number of ways to address this:
1. Somehow make the internal `update` work even with functions that might modify the log as a side-effect.
2. Do not expose higher-order operations `Xt.update` and `Xt.modify`  to users.
3. Document that the function passed to `Xt.update` and `Xt.modify` must not modify the transaction log and `assert` that inside the (internal) implementation of update.

For practical reasons I think that option 3 is the way to go.  The reason for that is that the "ideal fix" to make the internal `update` work even for cases where the log is mutated is likely to be expensive &mdash; in the general case it basically requires doing two splays on the log and avoiding that is the whole point of having those in the first place.  On other hand, the higher-order functions are quite important for performance reasons (i.e. for reducing accesses to the log), so it would be unfortunate to not give those to users.  Adding an assert to check that the `xt.cass` remains unmodified should be relatively cheap (and should get compiled away when `assert`s are disabled).

gasche:
If you want to provide semi-guarantees-by-convention that the higher-order argument does not mutate the transaction log, one way to do this would be the following:

1. Document that a local function should never capture a `xt` log variable. If the function uses `xt` in its body, it should take it as an `~xt` parameter. (Scala has an experimental feature called "capability" for this sort of should-not-be-captured values, and Jane Street's `local` stuff would also prevent that.) (You probably already need this convention to for correctness, independently of the `update`/`modify` business, otherwise an `xt` value could be leaked in a `unit -> unit` closure.)

2. In the higher-order functions, pass an extra `unit` parameter with the convention to rebind it to `xt` to shadow the ambient `xt` variable: `Xt.modify ~xt h (fun xt hv -> ...)`, where the inner `xt` has type `unit` (or a local type `no_access` or whatever to give even more suggestive error messages.)

polytypic:
> Document that a local function should never capture a `xt` log variable.

But that is not really a problem.

It is entirely fine to capture the log and it is one of the main advantages of the API.  You can, for example, use higher-order collection operations like `List.iter` and pass them a closure that captures the log and updates it during the iteration.

The problem is with `Xt.update` and `Xt.modify` as they are optimized to only access the log once and assume that the given function does not mutate the log.  It is a limitation that could be easily "fixed":

```ocaml
let modify ~xt r f = set ~xt r (f (get ~xt r))
```

But doing so one of the main reasons for `modify` to exist would be lost (i.e. that it only accesses the log once).  There is nothing that you could do with such a "fixed" `modify` or `update` that you cannot already do (with `get` and `set`).  You would just lose the optimization.

> Scala has an experimental feature called "capability" for this sort of should-not-be-captured values, and Jane Street's local stuff would also prevent that.

Hmm... I haven't looked at the proposal in detail, but I got the impression from talking with Stephen Dolan that locals can actually be used inside closures &mdash;  maybe the locals proposal also allows one to give higher-order arguments such types that they cannot capture locals.  At any rate, being able to use locals from closures without passing them as arguments was interesting to me as F# (and .NET in general) allows one to pass a variable by reference (i.e. [`byref<_>` arguments](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/byrefs)), but in F# such references cannot be captured &mdash; you can only pass them as arguments.  I've used this feature of F# specifically to ensure these kinds of properties (cannot leak, avoid allocations).

> You probably already need this convention to for correctness, independently of the update/modify business, otherwise an xt value could be leaked in a unit -> unit closure.

It is possible to leak the log, but after the log has been attempted there is really nothing you can do with the log.  You can record accesses to the log, if you capture it, but you cannot use the log to mutate anything aside from the log itself.  That is guaranteed by the type system.

> In the higher-order functions, pass an extra unit parameter with the convention to rebind it to xt

I see the idea, but strictly speaking that doesn't guarantee anything and would make the API more cumbersome as then you'd always have to wrap the function.  So, for example, you couldn't just say `modify ~xt q.back (List.cons elem)`.

PR #33, opened 2023-03-01
Designing lock-free algorithms

polytypic:
This PR adds a new section to the documentation on [**_Designing lock-free algorithms with k-CAS_**](https://github.com/ocaml-multicore/kcas/tree/designing-lock-free-algorithms#designing-lock-free-algorithms-with-k-cas).

TODO:
- [x] explain that essentially the same advice applies to both `Tx` and `Xt` APIs
- [x] explain that basically all access functions consult the transaction log once
- [x] add a warning about use of `Loc` module inside transactions


README.md
```diff
@@ -511,6 +534,721 @@ will essentially perform a
...
+
+```ocaml
+# let insert ~xt ~lt h x =
+    let h1 = `Node (
+        Loc.make `Null,
+        Loc.make 1,
+        x,
+        Loc.make `Null
+      ) in
+    Xt.set ~xt h (merge ~xt ~lt h1 (Xt.get ~xt h))
```
lyrm, line 647: I  tried that : 
```ocaml 
let insert ~xt ~lt h x =
  let h1 = `Node (Loc.make `Null, Loc.make 1, x, Loc.make `Null) in
  Xt.modify ~xt h (fun hv -> merge ~xt ~lt h1 hv)
```
and it does not work as expected. The following test returns `[5]` instead of the whole list `[ 3; 1; 4; 1; 5 ]` : 
```ocaml
let test () =
  let lt = ( > ) in
  let a_heap = leftist () in
  Xt.commit
    { tx = (fun ~xt -> List.iter (insert ~xt ~lt a_heap) [ 3; 1; 4; 1; 5 ]) };
  Xt.commit { tx = (fun ~xt -> remove_all ~xt ~lt a_heap) }
```

Whereas this following equivalent works:   
```ocaml
let insert ~xt ~lt h x =
  let h1 = `Node (Loc.make `Null, Loc.make 1, x, Loc.make `Null) in
  let hv = Xt.get ~xt h in
  Xt.set ~xt h (merge ~xt ~lt h1 hv)
````

Note : `update` does not seem to work as well.

Comment (polytypic) [reply README.md:647]: Thanks!  I added [a note on this to PR#30](https://github.com/ocaml-multicore/kcas/pull/30#issuecomment-1461950511).

File: README.md [hunk 2]
```diff
@@ -511,6 +534,721 @@ will essentially perform a
...
+      match Xt.exchange ~xt queue.middle [] |> List.rev with
+      | x :: xs ->
+        Xt.set ~xt queue.front xs;
+        Some x
+      | [] ->
+        match Xt.exchange ~xt queue.back [] |> List.rev with
+        | x :: xs ->
+          Xt.set ~xt queue.front xs;
+          Some x
+        | [] -> None
```
lyrm, line 998: I think more comments are needed here. Committing  a transaction inside a transaction feels pretty unnatural to me. I will try to explain why.

First, there is this idea that a transaction is something that is going to be reversed if it failed. But the internal committed transaction is a side effect that is not going to be reversed. 

Second, it seems to kind of defeat one of most interesting part of transactions : writing code in a sequential way, without considering what other domains are doing. It is very clear here : `back_to_middle` is called, and ones could expect `queue.middle` to not be empty since we just move `queue.front` in it. But, it could actually be empty because this operation is not part of the transaction. 

Also, this creates misleading code as this one: 
```ocaml
let test () =
  let switch_3_to_2 a =
    let tx ~xt =
      if Xt.get ~xt a = 3 then Xt.set ~xt a 2 in
    Xt.commit {tx}
  in
  let tx ~xt x  =
    Xt.set ~xt x 3;
    switch_3_to_2 x;
  in
  let v = Loc.make 0 in
  Xt.commit {tx = tx v};
  Loc.get v
````
that returns `3`. 


May be a useful comment for #30.

**Review by lyrm — COMMENTED** [review]
Partial review of README.

 I have read and tried everything except the last part (lock free hash table. I will review that later). 

This is very nice to read and very well explained.  

The important comment is the last one. I have concerns with the possibility of committing transaction inside a transaction (even if the example illustrates very well why this is useful).

File: README.md [hunk 3]
```diff
@@ -511,6 +534,721 @@ will essentially perform a
...
+interference or even starvation. For best performance it is important to keep
+transactions as short as possible. In particular, when possible, perform
+expensive computations after the transactions.
+
+Consider the following example of computing the size of a stack:
+
+```ocaml
+# let n_elems =
+    let tx ~xt =
+      Xt.get ~xt a_stack
```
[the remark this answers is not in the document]
polytypic: Good point!  I added a definition of `a_state` to the code block.

README.md
polytypic: > I think more comments are needed here.

Yes, I think that is fair observation.

The algorithm design technique described in this section is definitely more advanced than the fairly basic tips given in other sections.

> Committing a transaction inside a transaction [...]

It is actually a common pattern, or algorithm design technique, used in lock-free algorithms to _perform an atomic operation while preparing to perform another atomic operation_.

The emphasized part of the previous sentence can be argued to be a rephrase of what you said. A transaction (function) is something that can be run to prepare a specification of operations (the transaction log) to be attempted atomically (by the k-CAS-n-CMP algorithm).

(I wanted to highlight that part of what you said to draw the analogy (not trying to e.g. quote you out of context).)

> ...] feels pretty unnatural to me. I will try to explain why.

Yes, this is definitely an advanced technique, much like the whole of designing lock-free algorithms.

> First, there is this idea that a transaction is something that is going to be
> reversed if it failed.

Hmm... I would not say that a "transaction is [...] reversed if it failed", because that could give false impressions &mdash; Transactions aren't reversible and the construction of a transaction log does not (logically) mutate shared memory locations.

As a side note, I used "logically" above, because due to the underlying k-CAS-n-CMP algorithm, reading the content of a shared memory location may require the reader to help a potential writer to determine the content of the location being read. But I digress.

What the documentation says, for example, is that transactions are "attempted" and that they are "retried" in case of failure and the transaction logs are "constructed" by "record"ing operations.

> But the internal committed transaction is a side effect that is not going to
> be reversed.

Yes, that is basically analogous to the common technique of "helping" in lock-free algorithms.

In the Michael-Scott queue, for example, operations on the queue help each other to keep the tail updated. Updates of the tail do not change the logical state of the queue, but work towards making operations efficient and lock-free.

Similarly moving a non-empty back to the previously empty middle is a step that does not change the logical state of the queue, but is something that works towards that goal and multiple consumers can work together to make that happen.

The difference is that with **kcas** such helping operations are not limited to 1-CAS or single-word compare-and-set.

> Second, it seems to kind of defeat one of most interesting part of
> transactions : writing code in a sequential way, without considering what
> other domains are doing.

Well, yes and no.

- _Yes_ &mdash; When you use these kinds of lock-free algorithm design techniques, you make it more difficult to reason about the internal algorithmic details of transactions.

- _No_ &mdash; If you use the techniques correctly (making sure that the helping side-effects do not change the logical results) you can implement more fine-grained lock-free algorithms that can then be used compositionally without having to understand the internal details.

---

Thanks for raising these questions!

The technique obviously needed more fleshing out than what I had written before. As I've argued, the technique itself is not out of the ordinary &mdash; I would argue that it is based on common lock-free algorithm design techniques. It requires a good understanding of concurrency issues and the transaction mechanism and opens the possibility of writing more fine-grained and better performing lock-free algorithms. This is similar to what Reagents offer, but using the transaction mechanism. Furthermore, in a language like OCaml it would be difficult to completely prevent it, because OCaml makes it easy to perform side-effects anywhere.

I've added more elaboration on the topic to the documentation.

README.md
```diff
@@ -511,6 +534,721 @@ will essentially perform a
...
+[`exchange`](https://ocaml-multicore.github.io/kcas/doc/kcas/Kcas/Xt/index.html#val-exchange)
+we can reduce that to three:
+
+```ocaml
+# Xt.commit { tx = fun ~xt ->
+    let a_val = Xt.get ~xt a in
+    let b_val = Xt.exchange ~xt b a_val in
+    Xt.set ~xt a b_val }
+- : unit = ()
+```
```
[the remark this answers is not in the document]
polytypic: (Discussed in PR #30)

README.md
```diff
@@ -511,6 +534,721 @@ will essentially perform a
...
+    match h1, h2 with
+    | `Null, h2 -> h2
+    | h1, `Null -> h1
+    | (`Node (_, _, v1, _) as h1),
+      (`Node (_, _, v2, _) as h2) ->
+      let (`Node (h1l, npl, _, h1r) as h1), h2 =
+        if lt v1 v2 then h1, h2 else h2, h1 in
+      let l = Xt.get ~xt h1l in
+      if l == `Null then
+        Xt.set ~xt h1l h2
```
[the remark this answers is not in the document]
polytypic: It already is.  A leftist tree is left leaning &mdash; If the left branch `l` is `Null`, then the right branch must also be `Null`.  So, in that case the `npl` of the node (with the numbering used in the example) is already 1.

The algorithm is fairly faithfully transliterated from the book mentioned.  One difference is that in the book the npl of `Null` is defined to be `-1`.  In my transliteration I used `0` so the npl values are higher by one.  I had not looked at the wikipedia version before.  The merge operation in the book is written slightly differently (apparently to avoid some swaps and the redundant npl assignment).

README.md
lyrm: Thanks for the changes you made, it answers well my concerns !

lyrm:
Last part of my review : the `hashtbl` example is great. I saw nothing to change.

PR #34, opened 2023-03-02
Lockfreeonly locations

Dashy-Dolphin:


src/kcas.ml
```diff
@@ -174,8 +174,8 @@ let rec update_no_alloc backoff loc state set_after =
     let backoff = Backoff.once backoff in
     update_no_alloc backoff loc state set_after
 
-let is_obstruction_free casn =
-  Atomic.get casn == (Mode.obstruction_free :> status)
+let is_obstruction_free casn loc =
+  Atomic.get casn == (Mode.obstruction_free :> status) && loc.lock_mode == Mode.lock_free
```
polytypic, line 178: Should the comparison be against `Mode.obstruction_free`?

src/kcas.ml
```diff
@@ -189,8 +189,8 @@ let cas loc before state =
 module Loc = struct
   type 'a t = 'a loc
 
-  let make after =
-    { state = Atomic.make @@ new_state after; id = Id.get_unique () }
+  let make ?(mode_val = Mode.lock_free) after =
```
polytypic, line 192: I think `Mode.obstruction_free` should be the default.

polytypic: Also, the label could be just `mode`, i.e. `?(mode = Mode.obstruction_free)`.

Dashy-Dolphin: Will modify them accordingly.

src/kcas.ml
Dashy-Dolphin: Yes , It should be against Mode.obstruction_free.😅

PR #35, opened 2023-03-03
Lock-free mode argument to Locations

Dashy-Dolphin:
The Kcas operations are all done in either Lock-free mode or Obstruction-free mode based on the given value.
But , It is hypothesized that restricting specific locations to a certain mode only ( i.e., if an Kcas operation is operating in Obstruction-free mode but if the location is marked as Lock-free, then the operation on this  location moves forward  in Lock-free mode only) will provide us **significant speedup .**

Hence, This Pr Adds the argument to Location to **mark them as Lock-free or Obstruction-free.**


polytypic:
UGRC? 😸

src/kcas.ml
```diff
@@ -189,8 +190,12 @@ let cas loc before state =
 module Loc = struct
   type 'a t = 'a loc
 
-  let make after =
-    { state = Atomic.make @@ new_state after; id = Id.get_unique () }
+  let make ?(mode_val = Mode.obstruction_free) after =
```
polytypic, line 193: Maybe rename as `mode`?

Pretty much everything in OCaml is a value, so having `_val` as a suffix doesn't really add any information.

polytypic:
Suggestions:

- [ ] A getter for the mode of a location would be useful to have:

    ```ocaml
    module Loc : sig
      (* ... *)
      val get_mode : 'a t -> Mode.t
      (* ... *)
    ```

- [ ] A simple test that creates locations with different modes, verifies that the modes of the locations are as given and that the locations can be accessed (read and written) normally.

- [ ] A benchmark program that can be run with obstruction-free or with lock-free mode location and is designed to perform poorly with the obstruction-free location.  The intention of this is to validate that having the capability to mark a location as lock-free only makes sense.

Finally, using one full word for the mode is a bit pessimistic as only one bit of information is strictly needed.  One possibility would be to encode the information inside the id field in some manner.  Making the ids of lock-free locations negative, for example, would move them to the beginning in the order that locations are accessed during `determine` and that could also make it more likely for accesses of contentested locations more likely to succeed.

src/kcas.ml
Dashy-Dolphin: Ok😅...will modify accordingly.

src/kcas.ml
```diff
@@ -189,11 +190,16 @@ let cas loc before state =
 module Loc = struct
   type 'a t = 'a loc
 
-  let make after =
-    { state = Atomic.make @@ new_state after; id = Id.get_unique () }
+  let make ?(mode = Mode.obstruction_free) after =
+    {
+      state = Atomic.make @@ new_state after;
+      id = Id.get_unique ();
+      mode = mode;
```
polytypic, line 197: BTW, OCaml allows one to write this as just `mode`.  It is called [record field punning](https://dev.realworldocaml.org/records.html).

test/tx_loc_modes2.ml
```diff
@@ -0,0 +1,32 @@
+(* Location is in Lock-free mode
+   Each Thread increments the value at a and reads a in a loop*)
+
+module Loc = Kcas.Loc
+module Mode = Kcas.Mode
+module Tx = Kcas.Tx
+
+let a = Loc.make ~mode:Mode.lock_free 4
```
polytypic, line 8: You could combine the two programs into one by reading the mode from the command line arguments.

It would also be useful to read other parameters from the command line.

Something simple like this should do:

```ocaml
let loop_count = try int_of_string Sys.argv.(1) with _ -> 750
let mode = try if Sys.argv.(2) = "lock-free" then Some Mode.lock_free else None with _ -> None
(* ... *)
let a = Loc.make ?mode 4
```

test/tx_loc_modes2.ml
```diff
@@ -0,0 +1,32 @@
...
+  for _ = 1 to loop_count do
+    let alp = incr_loc a 1 in
+
+    let _ = Printf.printf "%d 1 i\n" @@ Tx.commit alp in
+
+    Unix.sleepf sleep_time
+  done
+
+let thread2 () =
+  for _ = 1 to loop_count do
```
polytypic, line 24: To make it more likely that the domains will be performing work simultaneously, you could use barrier synchronization.  Take a look at some of the other tests/benchmarks that use `Barrier` on how that can be done.

test/tx_loc_modes3.ml
```diff
@@ -0,0 +1,32 @@
+(* Location is in obstruction_free mode
+   Each Thread increments the value at a and reads a in a loop*)
+
+module Loc = Kcas.Loc
+module Mode = Kcas.Mode
+module Tx = Kcas.Tx
+
+let a = Loc.make ~mode:Mode.obstruction_free 4
+let read_loc loc = Tx.(return @@ Loc.get loc)
```
polytypic, line 9: I think I need to add warnings to the documentation on this, but unfortunately using operations from the `Loc` module is not how transactions should typically be created.  You should use the operations from the `Tx` module.

```diff
-let read_loc loc = Tx.(return @@ Loc.get loc)
+let read loc = Tx.get loc
```

`Loc.get loc` just immediately reads the value of `loc`.

`Tx.get loc`, on the other hand, returns a transaction that, when committed, reads the value of `loc` within the transaction.

There are situations when one might want to use `Loc.get` inside a transaction, but those are cases that need very careful analysis and understanding of how things work.

src/kcas.ml
Dashy-Dolphin: ok.

test/tx_loc_modes2.ml
Dashy-Dolphin: Will modify as to  use Barriers instead of making domains sleep.

src/kcas.ml
```diff
@@ -193,13 +192,17 @@ module Loc = struct
...
+        else (Id.get_unique () * 2) + 1);
     }
 
-  let get_id loc = loc.id [@@inline]
-  
-  let get_mode loc = loc.mode [@@inline]
+  let get_id loc = loc.id / 2 [@@inline]
+
+  let get_mode loc =
+    if loc.id mod 2 == 1 then Mode.lock_free else Mode.obstruction_free
```
polytypic, line 203: Unfortunately `x mod 2` gives a negative result if `x` is odd and negative.  `x land 1` compiles to a faster sequence of instructions.

Dashy-Dolphin: ohkk😅...will modify it to land.

test/tx_loc_modes2.ml
```diff
@@ -0,0 +1,40 @@
...
+let incr loc v =
+  Tx.(
+    let* v' = Tx.get loc in
+    Tx.set loc (v' + v))
+
+let thread1 () =
+  for _ = 1 to loop_count do
+    let alp = incr a 1 in
+
+    Barrier.await lock;
```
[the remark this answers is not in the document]
Dashy-Dolphin: Ok Now I get why they are performing equally well......Will Ideate on the interaction in interesting way.

src/kcas.ml
```diff
@@ -174,8 +174,8 @@ let rec update_no_alloc backoff loc state set_after =
     let backoff = Backoff.once backoff in
     update_no_alloc backoff loc state set_after
 
-let is_obstruction_free casn =
-  Atomic.get casn == (Mode.obstruction_free :> status)
+let is_obstruction_free casn loc =
+  Atomic.get casn == (Mode.obstruction_free :> status) && loc.id mod 2 == 0
```
polytypic, line 178: I think this should now be `loc.id > 0`.

polytypic:
I moved this work to PR #60 with some minor changes.  The commits in that PR should have the correct author information.

PR #37, opened 2023-03-19
Fix the lock-free update of Michael-Scott style queue tail

polytypic:
@lyrm, I realized that the protocol I devised for updating the tail of a Michael-Scott style queue was broken.  It suffered from pretty much the same kind of race condition as the previous attempts. 😞  So, this PR finally has a working solution to the Michael-Scott style queue tail leak.  The test in this PR was able to show that the previous attempts were wrong.  The logic of the new tail update algorithm is as follows:

> The tail update is only responsible for updating the tail as long as the new node added still has the true tail.  At any point some newly added node has the true tail, so this responsibility ensures that the tail gets updated.  To check whether we are still responsible, we compare: `get new_tail == Nil`.  If that is false, then someone else is responsible and we can stop.  The current `old_tail` must be read before that check.  If the `compare_and_set` fails, we must retry.  Otherwise we can stop as we've done our part.

At any rate, the take home lesson for me is to really stop trusting that my fine-grained concurrent code is correct unless I basically have a proof of it. 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -808,30 +808,28 @@ update the tail after the transaction has been successfully committed:
     let rec find_and_set_tail old_tail =
       match Xt.compare_and_swap ~xt old_tail Nil new_node with
       | Nil -> ()
-      | Node (_, old_tail) ->
-        find_and_set_tail old_tail in
-    let old_tail = Atomic.get tail in
-    find_and_set_tail old_tail;
-    Xt.post_commit ~xt @@ fun () ->
-    let rec fix_tail old_tail new_tail =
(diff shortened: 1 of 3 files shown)


polytypic:
> It may just be me but I had to read it multiple times to get it (I actually had to read the test to understand that I was reading it wrong). I would I got it more easily written as : _at least one of the queue is empty_.

Thanks!  I hadn't thought about this before, but my previous formulation "at most one ... non-empty" is much like using a double negative and those can be difficult to read.  I rephrased the invariant as "at least one of the queues is empty".

PR #38, opened 2023-03-26
Introduce `kcas_data` package of compositional lock-free data structures

polytypic:
This PR introduces a new `kcas_data` package of compositional lock-free data structures.

This first experimental version of `kcas_data` provides the following data structures:

* `Queue`
* `Stack`
* `Hashtbl`

The interfaces of these data structures are intentionally designed to be both subsets of `Stdlib` modules by the same names and to also to provide a subset of their operations as composable transactions.

This PR also includes an `Accumulator` that is more scalable than an `int Loc.t`, which is fundamentally non-scalable.

---

I'm glad that I decided to work on this PR, because I realized two important things:

1. The previous lock-free helping approaches I had written as examples were not correct. The problem was that in order to be able to perform lock-free helping steps, the locations involved must not already be in the log of the current transaction. To help with that I added the `is_in_log` operation to kcas. It is now used in the examples and also in this PR.

2. I realized that kcas has an inherent read anomaly issue that needs to be documented. When multiple locations are being updated atomically, it is possible for a reader to see values from different updates even though it is not possible to commit the transaction after having seen such inconsistent values.

   For example, suppose you have two locations `a` and `b` and that every update of those locations maintains that `a - b = 0`. A transaction that reads `a` and `b` may see that `a - b <> 0` and that anomaly can be surprising.

   Note that [MVCC](https://en.wikipedia.org/wiki/Multiversion_concurrency_control) and [TL II](https://link.springer.com/chapter/10.1007/11864219_14), for example, do not allow such anomalies.

Furthermore, the decision I made to try to mimic the `Stdlib.Hashtbl` interface to a very high degree and also keep all operations as strictly serializable turned out to be fruitful. The API of `Stdlib.Hashtbl` has not been designed for a lock-free implementation, which made the implementation much more challenging and required me to develop techniques to make that possible:

3. I devised lock-free, starvation avoiding, approaches to performing

   - rehashing/resizing,
   - snapshotting, and
   - in-place filtering and mapping

   of a hash table based on exposing the state of the operation to other domains.  It should also be fairly easy to improve the approach to allow multiple domains to work in parallel on different partitions of the buckets to complete the operations.  I intend to leave that to future work.

4. I realized the need to introduce a scalable `Accumulator` as the `Stdlib.Hashtbl` keeps a count of the number of bindings in a hash table and that does not scale if a single `int Loc.t` is used.  The current `Accumulator` implementation is naïve as it just uses `Domain.self () :> int` to index locations and it is easy to create domains in such patterns that an `Accumulator` performs poorly &mdash; although that would not be the expectation due to the costs of domains in multicore OCaml.  I also intend to leave possible improvements to `Accumulator` on this aspect to future work.

I did not implement `add_seq` or `replace_seq` operations. It should be possible to implement them using similar approaches as for the other "bulk" operations.   However, my assumption is that those operations are practically never needed to be strictly serializable and it kind of makes sense to have the user write `Seq.iter (fun (k, v) -> Hashtbl.add t k v) kvs` as it makes clear that bulk insert as a whole is not strictly serializable (only the individual `add`s are).  Of course, if the user needs `add_seq` or `replace_seq` to be strictly serializable in a high-contention scenario, it would make sense to provide those in the library.


src/kcas_data/accumulator.ml
```diff
@@ -0,0 +1,52 @@
...
+    | Some n_way -> n_way |> Int.min n_way_max |> Bits.ceil_pow_2
+  in
+  Array.init n_way (fun i -> Loc.make (if i = 0 then n else 0))
+
+let n_way_of = Array.length
+
+let get_self a =
+  let h = (Domain.self () :> int) in
+  (* TODO: Consider mixing the bits of [h] to get better distribution *)
+  Array.unsafe_get a (h land (Array.length a - 1))
```
polytypic, line 21: The problem here is that `Domain.self ()` does not make for a very good hash function and the bitwise AND will just discard any high order bits.  One potential solution would be to use a fast integer hash function here.   Probably a 16 or fewer bit hash function would be good enough for this purpose.

PR #48, opened 2023-04-11
Change to custom retry exception

polytypic:
This PR changes the library to introduce a custom `Retry` exception for signaling that a transaction should be retried.  Previously the [`Stdlib.Exit`](https://v2.ocaml.org/api/Stdlib.html#EXCEPTIONExit) exception was used for that purpose, but having a custom exception should make the intent clearer and also avoid surprises/issues where `Exit` might be used for other purposes.


polytypic:
It turns out that with blocking it makes sense to overhaul the retry interface even more.  So, I'm closing this PR.  The changes to retry will be in the blocking PR #32.

PR #68, opened 2023-05-10
Document assumption that (non-)blocking tx has not made updates

polytypic:


polytypic:
Closing this PR as PR #70 will offer better solutions.

PR #137, opened 2023-11-28
Use `Multicore_magic.domain_hash` in `Kcas_data.Accumulator`

polytypic:
This depends on [Add `domain_hash`](https://github.com/ocaml-multicore/multicore-magic/pull/9) and should not be merged before a new version of `multicore-magic` has been published.


polytypic:
I'm closing this as the approach of introducing an "instantaneous domain index" (#139) seems strictly preferable.

PR #139, opened 2023-11-30
Use `instantaneous_domain_index` in `Accumulator`

polytypic:
This PR changes `Accumulator` to use `instantaneous_domain_index` and make the accumulator grow automatically, which eliminates the need to have the optional `n_way` argument.

This depends on [Add `instantaneous_domain_index`](https://github.com/ocaml-multicore/multicore-magic/pull/10) and should not be merged before a new version of `multicore-magic` has been published.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -37,7 +37,7 @@
   (domain-local-timeout
    (>= 1.0.0))
   (multicore-magic
-   (>= 2.0.0))
+   (>= 2.1.0))
   (domain_shims
    (and
     (>= 0.1.0)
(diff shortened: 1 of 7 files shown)


kcas.opam.template
```diff
@@ -1 +1,4 @@
 doc: "https://ocaml-multicore.github.io/kcas/doc/kcas/Kcas/"
+pin-depends: [
+  [ "multicore-magic.dev" "git+https://github.com/ocaml-multicore/multicore-magic#dd9d8928fba7ab5dedef612a96bf01bd17cbd275" ]
```
polytypic, line 3: This pin-depend needs to be eliminated before merging.

PR #191, opened 2024-02-06
Adjust COC contacts

polytypic:
I do not want to be a contact for COC violations.

Do we need to have three COC contacts?

If so, perhaps someone else volunteers and can be added as a contact.


Sudha247:
I don't think we need three contacts, two should be fine. @balat, what do you think?

